PR #3, opened 2022-05-11
Enable resizable BAR support

sjkelly:
Resizable BAR support is a PCIe extension that allows resizing a PCIe device's
mappable memory/register space (also referred to as BARs - after the Base
Address Register that sets up the region). An important use case are GPUs.
While data center GPUs, generally have BAR sizes that match the size of video
memory, consumer and workstation GPUs generally declare only have 256MiB worth
of BARs mapping GPU memory to maintain compatibility with 32bit operating
systems. However, for performance (particularly when using PCIe P2P), it is
desirable to be able to map the entirety of GPU memory, necessitating
resizable BARs.

However, while PCIe ReBAR has been a standard for more than 10 years,
it was ill used until a few years ago and thus support is lacking.
On very recent motherboards (generally after 2020), a BIOS update might
be available that causes the firmware to read and reserve space for a
ReBAR expansion (or even perform the BAR resize itself). However, older
motherbards do not have this support.

Fortunately for us, the Linux kernel has some support to do its own PCIe
enumeration without relying on the firmware to do everything. Linux even has
support for resizable BARs, though in practice there are a number of important
limitations:

* There is currently no support for movable BARs in Linux. This means that if
  there are adjacent address space allocations, it is quite possible for BAR
  resizing to fail. There was a WIP patch series to resolve this issue at
  https://patchwork.ozlabs.org/project/linux-pci/cover/20201218174011.340514-1-s.miroshnichenko@yadro.com/
  but it appears to have faded out.

### DIFF
--- a/kernel-open/nvidia/nv-pci.c
+++ b/kernel-open/nvidia/nv-pci.c
@@ -289,10 +289,8 @@ resize:
            This is unlikely (because we had space before), but can happen. */
         nv_printf(NV_DBG_ERRORS, "NVRM: FATAL: Failed to re-allocate BAR1.\n");
         ret = -ENODEV;
-        goto done;
     }
 
-done:
     /* Re-enable memory decoding */
     pci_write_config_word(pci_dev, PCI_COMMAND, cmd);
 
--- a/kernel-open/nvidia/nv-pci.c
+++ b/kernel-open/nvidia/nv-pci.c
@@ -222,7 +222,6 @@ static void nv_init_dynamic_power_management
 
 
 
-
 static int nv_resize_pcie_bars(struct pci_dev *pci_dev) {
     struct pci_host_bridge *host;
     u16 cmd;
@@ -230,7 +229,12 @@ static int nv_resize_pcie_bars(struct pci_dev *pci_dev) {
     int ret = 0;
 
     // Check if BAR1 has PCIe rebar capabilities
+#if defined(NV_PCI_REBAR_GET_POSSIBLE_SIZES_PRESENT)
     u32 sizes = pci_rebar_get_possible_sizes(pci_dev, NV_GPU_BAR1);
+#elif
+        nv_printf(NV_DBG_INFO, "NVRM: pci_rebar_get_possible_sizes not supported\n");
+    u32 sizes = 0;
+#endif
     if (sizes == 0) {
         /* ReBAR not available. Nothing to do. */
         return 0;
@@ -239,9 +243,23 @@ static int nv_resize_pcie_bars(struct pci_dev *pci_dev) {
     /* Try to resize the BAR to the largest supported size */
     requested_size = fls(sizes) - 1;
 
+    /* Save the current size, just in case things go wrong */
+    old_size = pci_rebar_bytes_to_size(pci_resource_len(pci_dev, NV_GPU_BAR1));
+
+    if (old_size == requested_size) {
+        nv_printf(NV_DBG_INFO, "NVRM: %04x:%02x:%02x.%x: BAR1 already at requested size.\n",
+            NV_PCI_DOMAIN_NUMBER(pci_dev), NV_PCI_BUS_NUMBER(pci_dev),
+            NV_PCI_SLOT_NUMBER(pci_dev), PCI_FUNC(pci_dev->devfn));
+        return 0;
+    }
+
     /* If the kernel will refuse us, don't even try to resize,
        but give an informative error */
+#ifdef NV_PCI_FIND_HOST_BRIDGE_PRESENT
     host = pci_find_host_bridge(pci_dev->bus);
+#else
+    nv_printf(NV_DBG_INFO, "NVRM: pci_find_host_bridge not supported\n");
+#endif
     if (host->preserve_config) {
         nv_printf(NV_DBG_INFO, "NVRM: Not resizing BAR because the firmware forbids moving windows.\n");
         return 0;
@@ -261,8 +279,6 @@ static int nv_resize_pcie_bars(struct pci_dev *pci_dev) {
     /* Release BAR3 - we don't want to resize it, it's in the same bridge, so we'll want to move it */
     pci_release_resource(pci_dev, NV_GPU_BAR3);
 
-    /* Save the current size, just in case things go wrong */
-    old_size = pci_rebar_bytes_to_size(pci_resource_len(pci_dev, NV_GPU_BAR1));
 
 resize:
     /* Attempt to resize BAR1 to the largest supported size */
@@ -271,6 +287,8 @@ resize:
     if (r) {
         if (r == -ENOSPC)
             nv_printf(NV_DBG_ERRORS, "NVRM: No address space to allocate resized BAR1.\n");
+        else if (r == -EOPNOTSUPP)
+            nv_printf(NV_DBG_ERRORS, "NVRM: BAR resize resource not supported.\n");
         else if (r)
             nv_printf(NV_DBG_ERRORS, "NVRM: BAR resizing failed with error `%d`.\n", r);
     }
--- a/kernel-open/nvidia/nv-pci.c
+++ b/kernel-open/nvidia/nv-pci.c
@@ -223,7 +223,81 @@ static void nv_init_dynamic_power_management
 
 
 
+static int nv_resize_pcie_bars(struct pci_dev *pci_dev) {
+    struct pci_host_bridge *host;
+    u16 cmd;
+    int r, old_size, requested_size;
+    int ret = 0;
 
+    // Check if BAR1 has PCIe rebar capabilities
+    u32 sizes = pci_rebar_get_possible_sizes(pci_dev, NV_GPU_BAR1);
+    if (sizes == 0) {
+        /* ReBAR not available. Nothing to do. */
+        return 0;
+    }
+
+    /* Try to resize the BAR to the largest supported size */
+    requested_size = fls(sizes) - 1;
+
+    /* If the kernel will refuse us, don't even try to resize,
+       but give an informative error */
+    host = pci_find_host_bridge(pci_dev->bus);
+    if (host->preserve_config) {
+        nv_printf(NV_DBG_INFO, "NVRM: Not resizing BAR because the firmware forbids moving windows.\n");
+        return 0;
+    }
+
+    nv_printf(NV_DBG_INFO, "NVRM: %04x:%02x:%02x.%x: Attempting to resize BAR1.\n",
+        NV_PCI_DOMAIN_NUMBER(pci_dev), NV_PCI_BUS_NUMBER(pci_dev),
+        NV_PCI_SLOT_NUMBER(pci_dev), PCI_FUNC(pci_dev->devfn));
+
+    /* Disable memory decoding - required by the kernel APIs */
+    pci_read_config_word(pci_dev, PCI_COMMAND, &cmd);
+    pci_write_config_word(pci_dev, PCI_COMMAND, cmd & ~PCI_COMMAND_MEMORY);
+
+    /* Release BAR1 */
+    pci_release_resource(pci_dev, NV_GPU_BAR1);
+
+    /* Release BAR3 - we don't want to resize it, it's in the same bridge, so we'll want to move it */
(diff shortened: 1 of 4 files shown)


aritger:
Thanks for the contribution, sjkelly, and congratulations on the first pull request in this repository :)  I need to catch up a bit on the history of resizable BAR support within NVIDIA, and either I or someone else from NVIDIA will get back to this change soon.  Thanks again!

TheRawMeatball:
The use of `goto` is concerning to me.

Zanthed:
You have to adapt your code to the current codebase. Rewriting such a thing is for a separate time not relevant to this PR.

TheRawMeatball:
I am not advocating for a rewrite, but for not introducing more `goto` in new code.

Alcaro:
Yes, this patch adds 3 gotos.

There are 30 gotos in the file already, and Linus himself says gotos are good kernel coding style. I see no problem. https://web.archive.org/web/20130521051957/http:/kerneltrap.org/node/553/2131

Zanthed:
@TheRawMeatball Look at the current file. It's how the codebase was written. Removing the usage of goto is not relevant to this PR.

Keno:
One of the gotos is not required, since we dropped one of the error paths during rebase, so @sjkelly will remove that. But yes, other than that, gotos are fairly idiomatic in Linux. Much of the hate against gotos is based on the Dijkstra article, but there's some missing context there. Dijkstra was not really arguing against local gotos, but against global gotos that jump all over the place. Regardless, may I suggest not turning this PR into a discussion on the merits of gotos to allow the NVIDIA folks space for a proper review :)?

RustyRaptor:
The worst part about using a GOTO is how other programmers will react to it.

jackpot51:
I have a feeling there are firmware edgecases where an SMM handler or ACPI code could be holding on to the old BAR address. This kind of code is common for NVIDIA laptops and can vary between vendors or even models.

Keno:
> I have a feeling there are firmware edgecases where an SMM handler or ACPI code could be holding on to the old BAR address. This kind of code is common for NVIDIA laptops and can vary between vendors or even models.

Yes, it's possible, though in that situation, the firmware is supposed to ask the kernel not to move the PCIe BARs: https://github.com/torvalds/linux/blob/f400bea2d44beec76f7e7f45e5372ef790336a4d/drivers/acpi/pci_root.c#L919-L927. This patch respects that setting. I'm also not sure whether there are currently any laptop GPUs that include ReBAR support, since that's quite a new feature.

jackpot51:
> > I have a feeling there are firmware edgecases where an SMM handler or ACPI code could be holding on to the old BAR address. This kind of code is common for NVIDIA laptops and can vary between vendors or even models.
> 
> Yes, it's possible, though in that situation, the firmware is supposed to ask the kernel not to move the PCIe BARs: https://github.com/torvalds/linux/blob/f400bea2d44beec76f7e7f45e5372ef790336a4d/drivers/acpi/pci_root.c#L919-L927. This patch respects that setting. I'm also not sure whether there are currently any laptop GPUs that include ReBAR support, since that's quite a new feature.

It is good there is a way for firmaare to communicate this case. Yes, there are laptops that support resizeable bar, it became standard with RTX 3000 I believe.

ErickHoward:
> > I have a feeling there are firmware edgecases where an SMM handler or ACPI code could be holding on to the old BAR address. This kind of code is common for NVIDIA laptops and can vary between vendors or even models.
> 
> Yes, it's possible, though in that situation, the firmware is supposed to ask the kernel not to move the PCIe BARs: https://github.com/torvalds/linux/blob/f400bea2d44beec76f7e7f45e5372ef790336a4d/drivers/acpi/pci_root.c#L919-L927. This patch respects that setting. I'm also not sure whether there are currently any laptop GPUs that include ReBAR support, since that's quite a new feature.

Most 30 series laptops support resizeable BAR as far as I'm aware, I know that the Omen 15 supports it but usually I don't think it's visible in the UEFI like on desktop UEFIs.

divinity76:
could easily replace the gotos with a `for(;;){...continue;...break;}` , BUT that might arguably result in uglier code than the goto.

dmitriyc-nv:
Hi @sjkelly, if I understand the limitation of this patch correctly, it will only resize BAR1 if there is already enough address space available to do so. In other words, we cannot change the address space window that is already allocated to the GPU and can only operate within it. Can you please share a system configuration where these conditions exist and the patch successfully resizes BAR1?

Keno:
> Hi @sjkelly, if I understand the limitation of this patch correctly, it will only resize BAR1 if there is already enough address space available to do so.

Yes, this is correct.

> In other words, we cannot change the address space window that is already allocated to the GPU and can only operate within it.

No, this is not. The resources allocated to the GPU (BAR1 and BAR3) are released. The restriction is that address space needs to be available within the PCIe bridge that's upstream of the GPU. If there are no other devices in this PCIe bridge or the upstream PCIe bridge is the root complex this is not an issue, because the kernel will resize it (well, there's an issue about the size of the root complex window on AMD CPUs, but Linux already has code to open an additional huge window on boot to support rebar on AMD GPUs).

The mentioned restriction comes up when the PCIe topology includes a switch between the root complex and the GPU (pretty common on motherboards) and there is another PCIe device attached to the same switch. In that case, Linux currently does not have the ability to move the other device's PCIe BARs in order to resize the bridge window, which will likely cause the resize of the GPU BAR to fail.

> Can you please share a system configuration where these conditions exist and the patch successfully resizes BAR1?

We have successfully tested this on a number of systems. The one that @sjkelly has locally is an MBD-C9Z390-CG-O motherboard with an RTX A5000 plugged into the closest PCIe slot, with the other PCIe slot on the same PCIe switch left empty.

kernel-open/nvidia/nv-pci.c
```diff
@@ -223,7 +223,79 @@ static void nv_init_dynamic_power_management
...
 
 
+static int nv_resize_pcie_bars(struct pci_dev *pci_dev) {
+    struct pci_host_bridge *host;
+    u16 cmd;
+    int r, old_size, requested_size;
+    int ret = 0;
+
+    // Check if BAR1 has PCIe rebar capabilities
+    u32 sizes = pci_rebar_get_possible_sizes(pci_dev, NV_GPU_BAR1);
```
dmitriyc-nv, line 233: pci_rebar_get_possible_sizes() was added in v5.12 by commit 8fbdbb66f8c10. In order to avoid breaking the build on earlier kernels, a test for this function should be added to conftest.sh.

The kernel function has to be called without any arguments for the test to work, which is counterintuitive at first. The test should look something like this:
```
pci_rebar_get_possible_sizes)
            #
            # Determine if the pci_rebar_get_possible_sizes() function is present.
            #
            # Added by commit 8fbdbb66f8c10 ("PCI: Add resizable BAR infrastructure
            # ") in v5.12
            #
            CODE="
            #include <linux/pci.h>
            void conftest_pci_rebar_get_possible_sizes(void) {
                pci_rebar_get_possible_sizes();
            }"

            compile_check_conftest "$CODE" "NV_PCI_REBAR_GET_POSSIBLE_SIZES_PRESENT" "" "functions"
        ;;
```
It should also be added to nvidia.Kbuild:
```
NV_CONFTEST_FUNCTION_COMPILE_TESTS += pci_rebar_get_possible_sizes
```
After this, #if defined(NV_PCI_REBAR_GET_POSSIBLE_SIZES_PRESENT) can be used to either build this function, or make it a no-op for unsupported kernels (with possibly a helpful info or debug message).

kernel-open/nvidia/nv-pci.c
```diff
@@ -223,7 +223,79 @@ static void nv_init_dynamic_power_management
...
+    if (sizes == 0) {
+        /* ReBAR not available. Nothing to do. */
+        return 0;
+    }
+
+    /* Try to resize the BAR to the largest supported size */
+    requested_size = fls(sizes) - 1;
```
dmitriyc-nv, line 240: This should check current size and return success if it matches requested_size. Otherwise we would be reallocating BARs with each driver load even though they may already be resized.

kernel-open/nvidia/nv-pci.c
```diff
@@ -223,7 +223,79 @@ static void nv_init_dynamic_power_management
...
+
+    /* If the kernel will refuse us, don't even try to resize,
+       but give an informative error */
+    host = pci_find_host_bridge(pci_dev->bus);
```
dmitriyc-nv, line 244: pci_find_host_bridge() was merged in v5.15 by commit 59dc33252ee77, so this needs another conftest.sh entry. In this case, the implementation of pci_resize_resource() already checks host->preserve_config, so on kernels where this function is not exported I think we can safely skip this check here by surrounding it with conftest ifdefs.

kernel-open/nvidia/nv-pci.c
```diff
@@ -223,7 +223,79 @@ static void nv_init_dynamic_power_management
...
+    pci_release_resource(pci_dev, NV_GPU_BAR3);
+
+    /* Save the current size, just in case things go wrong */
+    old_size = pci_rebar_bytes_to_size(pci_resource_len(pci_dev, NV_GPU_BAR1));
+
+resize:
+    /* Attempt to resize BAR1 to the largest supported size */
+    r = pci_resize_resource(pci_dev, NV_GPU_BAR1, requested_size);
 
+    if (r) {
```
dmitriyc-nv, line 271: pci_resize_resource() can return -ENOTSUPP, and we shouldn't print a generic error in that case as it could happen quite often. In general I don't think we should print errors unless something failed catastrophically. Perhaps an info or debug print is sufficient here?

kernel-open/nvidia/nv-pci.c
```diff
@@ -223,7 +223,79 @@ static void nv_init_dynamic_power_management
...
+        else if (r)
+            nv_printf(NV_DBG_ERRORS, "NVRM: BAR resizing failed with error `%d`.\n", r);
+    }
+
+    /* Re-attempt assignment of PCIe resources */
+    pci_assign_unassigned_bus_resources(pci_dev->bus);
+
+    if ((pci_resource_flags(pci_dev, NV_GPU_BAR1) & IORESOURCE_UNSET) ||
+        (pci_resource_flags(pci_dev, NV_GPU_BAR3) & IORESOURCE_UNSET)) {
+        if (requested_size != old_size) {
```
dmitriyc-nv, line 283: Does it make sense to step through the available sizes down to the original size here?

kernel-open/nvidia/nv-pci.c
```diff
@@ -434,6 +506,12 @@ nv_pci_probe
         goto failed;
     }
 
+    if (nv_resize_pcie_bars(pci_dev)) {
```
dmitriyc-nv, line 509: Since this path runs every time .probe is called, we should add this feature behind a regkey module param at first (see nvrm_registry.h). Once we have good test coverage and feel confident that the feature is robust, we can enable this feature by default.

dmitriyc-nv (changes requested):

PAR2020:
Tracking internally via bug 3642641

dmitriyc-nv:
There has been no activity on this PR for over a month. @Keno, @sjkelly - do you intend to continue working on this patch?

sjkelly:
Yes I intend to address the comments unless someone is willing to expedite it internally.

kernel-open/nvidia/nv-pci.c
sjkelly: Do you mean `nv-reg.h`? Those expose kernel module options?

kernel-open/nvidia/nv-pci.c
sjkelly: I think so.

Keno:
@dmitriyc-nv Sorry, I lost track of this. I see @sjkelly did do the rebase, but it grew conflicts again. I can do another round of rebasing here if you would like.

dmitriyc-nv:
No worries, I've continued working on the patch internally and it is going through review.

Keno:
Excellent, thank you. Looking forward to it.

dmitriyc-nv:
A version of this patch has been merged in 530.30.02. Resizing can be enabled with the following module parameter:

NVreg_EnableResizableBar=1

PR #10, opened 2022-05-11
Fix capitalization in SECURITY.md

bigswag420:
:trollface:

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,8 +1,8 @@
 # Report a Security Vulnerability
 
 To report a potential security vulnerability in any NVIDIA product, please use either:
-* this web form: [Security Vulnerability Submission Form](https://www.nvidia.com/object/submit-security-vulnerability.html), or
-* send email to: [NVIDIA PSIRT](mailto:psirt@nvidia.com)
+* This web form: [Security Vulnerability Submission Form](https://www.nvidia.com/object/submit-security-vulnerability.html), or
+* Send email to: [NVIDIA PSIRT](mailto:psirt@nvidia.com)
 
 **OEM Partners should contact their NVIDIA Customer Program Manager**
 


SECURITY.md
```diff
@@ -1,8 +1,8 @@
 # Report a Security Vulnerability
 
 To report a potential security vulnerability in any NVIDIA product, please use either:
-* this web form: [Security Vulnerability Submission Form](https://www.nvidia.com/object/submit-security-vulnerability.html), or
-* send email to: [NVIDIA PSIRT](mailto:psirt@nvidia.com)
+* This web form: [Security Vulnerability Submission Form](https://www.nvidia.com/object/submit-security-vulnerability.html), or
```
mmaneetsingh, line 4: My thought is that line 4 and 5 here are continuation of the same sentence in line 3, hence first letter of these lines should not be capitalized. Line 3 would be too long so is broken down into bullets for ease of representation. This usage of bullets is very different from the usage of bullets in lines 10-14.

mmaneetsingh (commented):
Thank you bigswag420 for providing this change.

SECURITY.md
aritger: Sorry, @mmaneetsingh: I merged this before I saw your feedback.  I don't think the capitalization is big deal here, either way.  I've also ready merged this, so let's just leave it at that, unless you feel strongly?

mmaneetsingh: No worries. I agree with you it's not a big deal, so don't feel strongly about it.

aritger:
Thanks for the submission, @bigswag420.  We've applied your change.  Thanks.

BrunoGamerGH:
:trollface:

AntonUden:
:trollface:

THEGOLDENPRO:
:trollface:

TheNolle:
:trollface:

PR #13, opened 2022-05-11
Remove the template part from the license

Yardanico:
Stuff after "END OF TERMS AND CONDITIONS" explains how to apply GPL license to software, so it's not needed in the actual LICENSE files. 

See for example [qBittorrent](https://github.com/qbittorrent/qBittorrent/issues/7749) who had a similar issue of leaving out the template parts of GPL in the license. They [fixed it](https://github.com/qbittorrent/qBittorrent/commit/99341409f4cb6be54f506ddd5d7638ab4aec3ee7) by removing the template part, so I did the same here.


xfix:
GNU recommends leaving instructions in the license: https://www.gnu.org/licenses/gpl-faq.html#GPLOmitPreamble.

Yardanico:
@xfix okay

gradator:
The change was [reverted](https://github.com/qbittorrent/qBittorrent/commit/53cc8e11156d891e10f0315e98c8ddc8653fe6e7) in qBittorrent.

PR #16, opened 2022-05-11
Fix various typos in nvidia/src/kernel

alexisgeoffrey:


mmaneetsingh (approved):
Thank you alexisgeoffrey for providing this change.
I verified, this change looks good. We will incorporate this in our next release.

AlexFolland:
Will you be merging pull requests here and developing in the open, or only committing one release at a time?

aritger:
> Will you be merging pull requests here and developing in the open, or only committing one release at a time?

Because we will be sharing this code with our proprietary driver, we won't be developing in the open for now.  So far, our strategy is to apply proposed changes to our internal code base, merge pull requests on github, and then do one NVIDIA github commit per driver release (and because the internal code base also contains the change, the release-time commit should not revert the merged pull request).  It is not a great workflow, but we're trying to navigate the constraints as best we can.

PR #22, opened 2022-05-12
Enforce C standard compliance and fix ACPI null pointer dereference

amada95:
Simple fix for a minor compiler warning to dip my toes into this codebase.

### DIFF
--- a/kernel-open/nvidia/nv-acpi.c
+++ b/kernel-open/nvidia/nv-acpi.c
@@ -737,6 +737,9 @@ void NV_API_CALL nv_acpi_methods_uninit(void)
 #if defined(NV_ACPI_BUS_GET_DEVICE_PRESENT)
     acpi_bus_get_device(nvif_parent_gpu_handle, &device);
 
+    if (device == NULL)
+        return;
+        
     nv_uninstall_notifier(device->driver_data, nv_acpi_event);
 #endif
 
--- a/kernel-open/nvidia-drm/nvidia-drm-crtc.c
+++ b/kernel-open/nvidia-drm/nvidia-drm-crtc.c
@@ -1141,11 +1141,13 @@ void nv_drm_enumerate_crtcs_and_planes(
         }
 
         for (layer = 0; layer < pResInfo->numLayers[i]; layer++) {
+            struct drm_plane *overlay_plane;
+
             if (layer == NVKMS_KAPI_LAYER_PRIMARY_IDX) {
                 continue;
             }
 
-            struct drm_plane *overlay_plane =
+            overlay_plane =
                 nv_drm_plane_create(nv_dev->dev,
                                     DRM_PLANE_TYPE_OVERLAY,
                                     layer,


Det87:
Is C90 still relevant?

baryluk:
NACK

Conforming strictly to C90 makes code less readable.

Leave it as is and just bump flags to the compiler to use `gnu99`.

jackghg:
pointless

amada95:
I agree that c99 is a much better, more readable and more relevant standard, but I'd guess that the standard NVIDIA is using here was a deliberate choice so this PR just falls in line.

jackghg:
speachless

amada95:
This PR now also addresses issue #109

ryao:
I feel like there should be 1 PR per issue. Also, most projects do not like to see merge commits in PRs. Each logical change should be its own commit, but merge commits just document how things were put into the git repository, which make them clutter. I suggest rebasing, cherry-picking and force pushing to clean up the PR.

amada95:
> I feel like there should be 1 PR per issue. Also, most projects do not like to see merge commits in PRs. Each logical change should be its own commit, but merge commits just document how things were put into the git repository, which make them clutter. I suggest rebasing, cherry-picking and force pushing to clean up the PR.

Thanks for the friendly advice, I'm relatively new to open-source contributing and it's nice to be given some constructive advice. I went ahead and cleaned up the merge commit, but I don't necessarily want this PR to get drowned out so I won't be splitting the commits into separate PRs (however that's duly noted for the future, I know it's not ideal for the maintainers).

mtijanic:
For context, our guidelines (which we hope to publish soon) generally say that we should prefer C90 in the general case due to some UB related to C99 mixed declarations, but strict C90 compatibility is not a goal and it is okay to use C99 style declarations when there is a strong justification to deviate from the guidelines.

Also, for what it's worth, some of the platforms we support from the shared codebases will accept most C99 features but not `for (int i;;)`, so this one is avoided more strictly.

amada95:
> For context, our guidelines (which we hope to publish soon) generally say that we should prefer C90 in the general case due to some UB related to C99 mixed declarations, but strict C90 compatibility is not a goal and it is okay to use C99 style declarations when there is a strong justification to deviate from the guidelines.
> 
> Also, for what it's worth, some of the platforms we support from the shared codebases will accept most C99 features but not `for (int i;;)`, so this one is avoided more strictly.

Thanks for clarifying. I made this change initially based on the C90 standard, but it was also consistency with other functions within the file and as such makes sense to implement for coherence reasons as well.

amada95:
Hello, just checking in - any updates? Normally I wouldn't bump an issue like this, but there have been several PRs reviewed since this one.

kernel-open/nvidia/nv-acpi.c
```diff
@@ -737,6 +737,9 @@ void NV_API_CALL nv_acpi_methods_uninit(void)
 #if defined(NV_ACPI_BUS_GET_DEVICE_PRESENT)
     acpi_bus_get_device(nvif_parent_gpu_handle, &device);
 
+    if (device == NULL)
```
tallossos, line 740: Why aren't you using the `ACPI_FAILURE` macro to check the result of `acpi_bus_get_device`?
You could check other places where `acpi_bus_get_device` is being used and do the same.

mtijanic:
As @ryao mentioned, it is best to not group unrelated changes into a single PR. Particularly if they also touch different kernel modules, as these are often maintained by different people and have to be merged separately anyway.

My advice is to close this and open new individual PRs per issue (or class of issues) that you want to address. Keeping in mind that cosmetic stuff will just be tagged appropriately and deferred for now.

By the way, this `device` null deref is only partially fixed, as `NV_ACPI_BUS_GET_DEVICE_PRESENT` could be undefined in which case your nullcheck is skipped.

amada95:
Alright, thank you all for the explanation and your time.

jackghg:
kernel just moved to c11... just to let you know...

PR #26, opened 2022-05-12
Update README.md, Added Requirement section for linux header files

GERZAC1002:
Added Requirements section as the linux header files are required and not installed by default on every Linux system.
Also added link to the current version of the Beta driver 515.43.04


fzn0x:
Hi, @GERZAC1002 can you make this PR add the requirement section only and revert the link on the How to Build section? I took the initiative to make knowledge of tags efficient and not changed every time there is a new version.

Thank you,

aritger:
Thanks for the suggestion, but I'm reluctant to list specific required packages, because they will vary by Linux distribution.

PR #29, opened 2022-05-12
docs: introduce code-block highlight

KisaragiEffective:


pranavo72bex (approved):

aritger:
Thanks for the suggestion, but I'm happy enough with the current formatting of these commands in README.md, so I'm inclined to leave this as-is.  I'm going to go ahead and close, but let me know if there is some benefit to code-block highlight (sorry, I don't know github's markdown that well, so I could be missing something).

KisaragiEffective:
@aritger
Thanks for your review! I've updated the description to explain benefits. I'm glad if you would re-consider this PR, thank you.

PR #31, opened 2022-05-12
run build on GitHub Actions

KisaragiEffective:


.github/workflows/build.yml
```diff
@@ -0,0 +1,22 @@
...
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
```
svenstaro, line 18: This is not needed in this repository.

KisaragiEffective, line 18:
```suggestion
```

KisaragiEffective, line 18:
```suggestion
```

KisaragiEffective: indeed, fixed in 7bab9b8.

.github/workflows/build.yml
```diff
@@ -0,0 +1,22 @@
...
+
+      - name: Build
+        run: |
+          make modules -j`nproc`
```
Docteh, line 20: I was looking through the other PR's and #48 is suggesting that -j`nproc` be replaced with -j$(nproc)

Can this be done here as well?

.github/workflows/build.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Build
+        run: |
+          make modules -j`nproc`
```
KisaragiEffective, line 20:
```suggestion
          make modules -j$(nproc)
```

.github/workflows/build.yml
KisaragiEffective: sure, done in ec11aac

.github/workflows/build.yml
```diff
@@ -0,0 +1,20 @@
+name: Build and deploy
```
PurHur, line 1: The name is misleading because its only build not deployed

KisaragiEffective, line 1:
```suggestion
name: Build
```

KisaragiEffective: right, fixed in 871674a

.github/workflows/build.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Build
+        run: |
+          make modules -j$(nproc)
```
Margen67, line 20:
```suggestion
        run: make modules -j$(nproc)
```

KisaragiEffective:
> This is only building on `amd64`. #111 is building on both `amd64` & `arm64` and installing on both as well.

@dvcastillo I believe this was resolved by 9851180.

.github/workflows/build.yml
```diff
@@ -9,11 +9,22 @@ on:
       - main
 
 jobs:
-  build:
-    runs-on: ubuntu-20.04
+  build-amd64:
+    runs-on: ubuntu-20.04
```
Margen67, line 13: Whitespace
```suggestion
    runs-on: ubuntu-20.04
```

.github/workflows/build.yml
```diff
@@ -9,11 +9,22 @@ on:
...
-  build:
-    runs-on: ubuntu-20.04
+  build-amd64:
+    runs-on: ubuntu-20.04          
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
-
-      - name: Build
+      - name: build on amd64
```
Margen67, line 17:
```suggestion
      - name: Build on amd64
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: actions/checkout@v3
+      - name: build on amd64
+        run: make modules -j$(nproc)
+  build-aarch64:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - uses: uraimo/run-on-arch-action@v2
+        name: build on aarch64
```
KisaragiEffective, line 25:
```suggestion
        name: Build on aarch64
```

This should be changed as well, right?

Margen67: Yes.

james-crowley:
@KisaragiEffective I am not sure if GitHub Actions offers Arm hardware but I would recommend against running via QEMU/emulation. 

QEMU is going to be slower than running on native hardware and when running low level code like this it can lead to false positives. 

I think running on native hardware is the way to go. Additionally, having a place where we can actually test the against real GPUs before cutting a release would be amazing!

PurHur:
To make ARM builds nvidia could sponsor a server for the runner. Also the base actions will need some computing so this has to be solved in the future. I guess nvidia already has some buildservers somewhere and could install a runner for that.

KisaragiEffective:
@james-crowley wrote:
> I am not sure if GitHub Actions offers Arm hardware
Self-hosted runners can be run on different architecture, but GitHub hosted-runner runs on [Dv2 and DSv2-series](https://docs.microsoft.com/azure/virtual-machines/dv2-dsv2-series#dsv2-series), which is (if I read it correct) on x86_64.

@james-crowley wrote:
> QEMU is going to be slower than running on native hardware and when running low level code like this it can lead to false positives.
> 
> I think running on native hardware is the way to go. Additionally, having a place where we can actually test the against real GPUs before cutting a release would be amazing!

> To make ARM builds nvidia could sponsor a server for the runner. Also the base actions will need some computing so this has to be solved in the future. I guess nvidia already has some buildservers somewhere and could install a runner for that.

It makes sense, but I'm prefer checking build earlier than runs natively in earlier point since there are several PRs already. I wouldn't like to assume its support, there's no guarantee to be provided.

.github/workflows/build.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Build on amd64
+        run: make modules -j$(nproc)
+  build-aarch64:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - uses: uraimo/run-on-arch-action@v2
```
KisaragiEffective, line 24:
```suggestion
      # NOTE: GitHub hosted runners does not support aarch. See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#cloud-hosts-for-github-hosted-runners for more info.
      - uses: uraimo/run-on-arch-action@v2
```

PurHur:
@nvidia @aaronp24 how is the progress? It would be really nice if the company would sponsor the needed build servers for this great project. Could you please escalate this topic to upper management? Thanks.

PR #32, opened 2022-05-12
Use STATE_MASK instead of MAP_MASK

ElijahPepe:
This commit fixes a TODO entry in addrtree.c to use STATE_MASK when referring to stateMask in _addrtreeNodeIndexHasSeeChildSet; only the state mask should matter in this case.


ElijahPepe:
Thank you for the quick response! I was able to validate that using STATE_MASK over MAP_MASK wouldn't cause serious issues using cuda_memtest. It would be helpful if more testing could be done on this, however.

mtijanic:
Thanks @ElijahPepe . I will check with the engineers maintaining this code and either bring them here or relay their advice. Stay tuned.

mtijanic:
Hi @ElijahPepe , I checked with the engineers maintaining this code and they said the whole Address Tree allocator is not enabled on Linux and there is no good test infrastructure in place that would validate this change right now.

Thanks for the PR, but I'm going to close it as it's dead code for this driver.

PR #34, opened 2022-05-12
Replaced unsafe functions with better alternatives

AntoonBeres:
removed some uses of strcpy and strcat in favor of `strncpy` and `strncat`, in favor of the safer `strncpy` and `strncat`

Ideally strcpy_s and strcat_s should be used, but don't know if that is possible with compatibility constraints etc .. no good reason to use strcpy and strcat when strncpy and strncat can be used however.


ElijahPepe:
`strncpy` is not a safe replacement for `strcpy` despite the unfortunately similar names and the former should never be used when dealing with a destination string that is zero-terminated. A safer replacement for `strcpy` is `strlcpy`, which is not a standard C function.

AntoonBeres:
> 

True, but at least it's a slight improvement over strcpy and unlike other safe alternatives it's part of standard C90. Unless you manually want to craft a function checking for memory safety (like strcpy_s or strlcpy). 

It's not perfect, but at least it's better than plain strcpy and strcat

a1batross:
>but at least it's a slight improvement over strcpy

It's not. You can't expect that strncpy will zero terminate.

Also:

>part of standard

You're writing kernel code. No standard library exists here, so you can safely use whatever you want: [strscpy](https://www.kernel.org/doc/htmldocs/kernel-api/API-strscpy.html) preferred. :)

kernel-open/nvidia/nv-caps.c
```diff
@@ -594,9 +594,9 @@ static nv_cap_t* nv_cap_alloc(nv_cap_t *parent_cap, const char *name)
         return NULL;
     }
 
-    strcpy(cap->path, parent_cap->path);
-    strcat(cap->path, "/");
-    strcat(cap->path, name);
+    strncpy(cap->path, parent_cap->path, len);
+    strncat(cap->path, "/", len);
+    strncat(cap->path, name, len);
```
gengkev, line 599: This code might be misunderstanding what the third argument of `strncat` is. If the intention is to have a line of code that can never write `len` bytes past the start of `cap->path`", then the correct code is `strcat(cap->path, name, len - strlen(cap->path) - 1);`. Almost certainly better to leave the code as it is...

AntoonBeres: My bad, will fix this when I get home, will reinstate strcat, I will also replace strncpy by by strscpy as suggested here before. 

in the files that are not linux-specific, i can first do a #ifdef linux, and keep using strncpy if the linux isn't defined and otherwise use the safer strscpy

N-R-K:
All the buffers are allocated properly and checked against NULL. There's nothing """unsafe""" going on in the current code. And funny enough, this PR itself is "unsafe" because the `strncat` calls are wrong.

AntoonBeres:
> 

True, the current usage was not unsafe, but the functions themselves are and it is bad practice to use them. Concerning my usage of strncat, you're right, that was my bad, fixed that now.

N-R-K:
> the current usage was not unsafe, but the functions themselves are

This mentality that some function are "unsafe" and using "safe" function is somehow magically going to solve any issue only gives a false sense of security and does more harm than good.

```c
char *s = malloc(strlen(d)+1);
assert(s != NULL);
strcpy(s, d);
//
char *s = malloc(strlen(d));
assert(s != NULL);
strscpy(s, d, strlen(d)+1);
```

And I'm not even against removing strcat calls since misuse of it has caused many issues in the past.

But anyone who thinks a function is "unsafe/safe" has a completely failed to grasp what the *actual* problem is and thus is going to repeat the history with the "safer" variants as well.

kernel-open/nvidia/nv-caps.c
ajmes: You could use strlcat instead of strcat:
https://linux.die.net/man/3/strlcat

aritger:
I agree with @N-R-K.  I don't immediately see a problem here that needs fixing.  If you see a specific bug that needs fixing, please spell out the scenario.  For now, I'll close this pull request.

sergiomb2:
may we need this pull request for kernel 7.2 since strncpy Is Finally Gone in kernel Linux 7.2-rc1
https://michealgarner.co.uk/blog/strncpy-is-finally-gone-what-linux-72-rc1-means-for-kernel-security

PR #37, opened 2022-05-12
Improve `conftest.sh` Script

Nitepone:
- Remove non-posix local usage
- Remove trailing whitespace
- Fix broken test caused by missing quotes

### DIFF
--- a/kernel-open/conftest.sh
+++ b/kernel-open/conftest.sh
@@ -628,10 +628,10 @@ compile_test() {
         set_pages_array_uc)
             #
             # Determine if the set_pages_array_uc() function is present.
-            # It does not exist on all architectures. 
-            # 
-            # set_pages_array_uc() was added by commit 
-            # 0f3507555f6fa4acbc85a646d6e8766230db38fc ("x86, CPA: Add 
+            # It does not exist on all architectures.
+            #
+            # set_pages_array_uc() was added by commit
+            # 0f3507555f6fa4acbc85a646d6e8766230db38fc ("x86, CPA: Add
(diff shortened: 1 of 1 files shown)


gghhkm (approved):

Luexa:
You are correct in that this usage of `local` is not POSIX-compliant. However, POSIX sh *does* include support for local variables, and it is good practice to use them!

Here is what `man dash` (sh implementation that is very strict about being POSIX compliant) has to say:

> Variables may be declared to be local to a function by using a `local` command.  This should appear as the first statement of a function, and the syntax is
> 
> ```sh
> local [variable | -] ...
> ```
> 
> `local` is implemented as a builtin command.
> 
> When a variable is made local, it inherits the initial value and exported and readonly flags from the variable with the same name in the surrounding scope, if there is one.  Otherwise, the variable is initially unset.  The shell uses dynamic scoping, so that if you make the variable `x` local to function `f`, which then calls function `g`, references to the variable `x` made inside `g` will refer to the variable `x` declared inside `f`, not to the global variable named `x`.

Essentially what you need to do is declare all local variables used in the function as the first statement of the function.

Example POSIX-compliant usage of `local` which also makes sure environment is not polluted with the values of these variables (by unsetting them, as that is the only way to unexport a shell variable):

```sh
example_function () {
    local    foo bar
    unset -v foo bar

    foo="abc"
    bar="xyz"
}
```

aritger:
Thanks for the contribution.  I've merged all three patches.
Sure enough, it looks like `local` isn't POSIX.  The usage of local in conftest is pretty minor, so just removing the use of `local` as done in the first two commits seems simplest.  Thanks.

PR #63, opened 2022-05-12
common/nvswitch: [fru_nvswitch] do not increment index before performing loop operation

rmnscnce:
This should:
- Take account the first byte of the data array into the checksum
- Prevent read overflows caused by the last iteration of the loop operation

P.S.: This commit is being made under the naive assumption of the data structure for `NvU8 *data`, and is done so because I have not found a use of this function inside the source archive. Kindly just reject this PR if it's actually intentional.


Joshua-Ashton (changes requested):
This commit does nothing.

Using prefix or postfix incrementation in that part of the for-loop does not change the result in any way.

rmnscnce:
Thanks for the feedback. I'll just close the PR, then.

PR #64, opened 2022-05-12
Fix capitalization in Makefile

QuantumCoded:


Joshua-Ashton:
The target is called `clean`.

shbaskaran:
Thanks for the contribution QuantumCoded. As pointed out by Joshua-Ashton, the comment block is referring to the target 'clean'. Arguably it can be capitalized in the comment block, but leaving this as-is. 

I hope you will choose to contribute other changes to our codebase!

PR #71, opened 2022-05-12
Several typos

sevenzing:
found some typos `¯\_(ツ)_/¯ `


niv:
Thank you for your interest! While those are valid typos and should be fixed, they'll eventually be addressed when someone touches that code for a bigger change. Doing a PR for every single typo is very noisy. Let's keep those to a minimum please.

PR #73, opened 2022-05-12
readme: give a reference point for the word "time-consuming"

jengelh:


ValZapod:
There is no any binaries in the repo... nvidia.ko to.

aritger:
This example reference point feels a bit ephemeral... build times will vary greatly based on the hardware used, and it could vary from one release to the next as code is added or removed.  I'm inclined to leave the wording vague as just "time-consuming".  Sorry.

jengelh:
Building the Linux kernel with almost-all-m takes 14 minutes, and building Chromium is like 3 hours. My point was that open-gpu-kernel-modules is anything _but_ time-consuming, both relative and absolute.

ValZapod:
> building Chromium is like 3 hours.

That is because C++ is used by Chromium too actively, linux kernel is pure C and tested by gcc devs.

jengelh:
>That is because C++ is used by Chromium too actively, linux kernel is pure C and tested by gcc devs.

Hardly. Chromium consists of somewhere like 40000, and Linux of maybe 30000 build targets (C++ or not). The NV module is **nowhere** near that. As such, it should not suggest that its own build is (overly) time-consuming.

PR #78, opened 2022-05-12
Default to nanosecond clock time with GPL nvidia driver

ad1rie1:
Hello, 

Get nanosecond présentation timestamp now nvidia is a GPL driver

Dear,

### DIFF
--- a/kernel-open/nvidia-uvm/uvm_linux.h
+++ b/kernel-open/nvidia-uvm/uvm_linux.h
@@ -305,7 +305,6 @@ static inline uint64_t NV_DIV64(uint64_t dividend, uint64_t divisor, uint64_t *r
 }
 #endif
 
-#if defined(CLOCK_MONOTONIC_RAW)
 /* Return a nanosecond-precise value */
 static inline NvU64 NV_GETTIME(void)
 {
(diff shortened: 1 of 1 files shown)


ValZapod:
Of god! BTW, windows recently also started exposing ns time (windows 7?), hope those are using ns.

mtijanic:
Hello, welcome and thank you for your interest. As you're probably aware, this file is part of the nvidia-uvm.ko driver that has already been published as dual GPL/MIT. This driver is used in conjunction with both the newly published open nvidia.ko as well as traditional closed source version.

Applying this patch would therefore break the closed source driver which cannot depend on the GPL symbols. If I'm reading the code correctly, on the new driver we are already using the nanosecond precision and the fallback only applies to the proprietary module. 
In other words, the issue is with the 'dead' code and not functional. Is my understanding here correct?

ad1rie1:
yes

ValZapod:
Wait, wait! You can just do a #define! Obviously!

ad1rie1:
@ValZapod the new driver already uses nanosecond precision as default, it's a mistake of myself.

ValZapod:
So there is a completly different behaviour in closed source driver that still uses microseconds? Wow, that is even more horrible :)

PR #89, opened 2022-05-12
Remove extra '.' and capitalize some comments

0xk0sta:
Please, this needs to be fixed ASAP.


Shadofer:
Why asap?

ValZapod:
Tell your OCD that thise are titles in most cases and those should have no full stop. On fact your commit name does not have a dot.

0xk0sta:
> Tell your OCD that those are titles in most cases and those should have no full stop. In fact your commit name does not have a dot.

Wow Thanks! I didn't notice that.

PR #91, opened 2022-05-12
Optimized for loop iterator

eljommys:
++i is more efficient than i++

### DIFF
--- a/src/common/shared/nvstatus/nvstatus.c
+++ b/src/common/shared/nvstatus/nvstatus.c
@@ -70,7 +70,7 @@ const char *nvstatusToString(NV_STATUS nvStatusIn)
     static NV_PRINTF_STRING_SECTION const char rm_pvt_UNKNOWN_str[] = "Unknown error code!";
     NvU32 i;
     NvU32 n = ((NvU32)(sizeof(g_StatusCodeList))/(NvU32)(sizeof(g_StatusCodeList[0])));
-    for (i = 0; i < n; i++)
+    for (i = 0; i < n; ++i)
     {
         if (g_StatusCodeList[i].statusCode == nvStatusIn)
         {


eljommys:
No

0xk0sta:
> No. It's not. The compiler does that itself. Both the original version and your version will result in exactly the same identical bytecode. This has already been discussed in #63

That's a fact.

Leseratte10:
> No

What do you mean "no"?
We are no longer in the assembly days of 1980. A standard gcc compiler will compile a standalone `i++` to the exact same bytecode as `++i`.

eljommys:
image

alcaparra:
Excellent, I compiled the module with this modification and it performed 26% faster. Benchmarks:

```
Master -> exec: 30ms, runtime: 49ms
Modified -> exec: 22ms, runtime: 38ms
```

eljommys:
> Excellent, I compiled the module with this modification and it performed 26% faster. Benchmarks:
> 
> ```
> Master -> exec: 30ms, runtime: 49ms
> Modified -> exec: 22ms, runtime: 38ms
> ```

Told ya

Joshua-Ashton:
I was going to comment that this PR does nothing -- but the rest of your PRs are just useless spam bullshit so I know you already know that.

alcaparra:
@Joshua-Ashton Please use respectful language, there is no need for cursing. Also, you crying is annoying everyone.

eljommys:
I actually compiled with this and it shows a similar result
github.com/m3t4f0r4/fastcc

ValZapod:
> No. It's not. The compiler does that itself.

In fact not every compiler, one from 20 years ago does not (Software Toolworks C80 compiler). See https://stackoverflow.com/questions/24886/is-there-a-performance-difference-between-i-and-i-in-c?rq=1

It is also true that this is a good style to use ++i in for loops (the only style just like *stuff in pointers) and that it may be faster if it is classes operation or in case of overloading.

The fact that it is faster means nothing: show us the assembly! If you do not know assembly then you cannot do such uoptimisations.

okawo80085:
Can someone reject this? This is literally useless.

dev11n:
If you're going to contribute, at least contribute something meaningful. `i++` vs `++i` in this case isn't even close to being meaningful.

NotMyWing:
This should be rejected, and both of these people should be banned from committing altogether.

kafteji:
oh man.. another green grid seeker :expressionless:

niv:
Thank you for your suggestion. The compiler handles this just fine by itself.

somerandomcloud:
Peak Github moment

team-orangeBlue:
g. uys  .m....... why am i the only ine giving the opposite reaction s tk all comments !????!!!!!!n!?1//1/1?2?1!!1 i need more friends helping me !!!!!!!!!!!!!!!!!!!!!!!!!!!!

PR #79, opened 2022-05-12
fix typo

DoraeM2112:
readability↑

PR #93, opened 2022-05-12
Remove extra whitelines

juanrd10:


leirons:
seems useless

juanrd10:
But it brings clarity, I deserve the merge.

Shinyzenith:
if you're gonna submit such a useless patch, might as well rebase and squash them into 1 commit with a sane commit message.

fastman:
> if you're gonna submit such a useless patch, might as well rebase and squash them into 1 commit with a sane commit message.

moar commit, moar fame

niv:
Thanks for the report. All the extra whitespace will be handled in the future. Also see #79, also see internal bug 3643422.

PR #101, opened 2022-05-12
Update HTTP link to HTTPS

alcaparra:
`http://www.nvidia.com` redirects to `https://www.nvidia.com`, so linking to http is reduntant.
More importantly, it could be considered a security issue, do to the nature of plain http. An attacker could perform a MITM attack and redirect the HTTP request to a malicious website.
```
> GET / HTTP/1.1
> Host: www.nvidia.com

Response:
< HTTP/1.1 307 Temporary Redirect
< Location: https://www.nvidia.com/
```

### DIFF
--- a/src/nvidia-modeset/src/nvkms-framelock.c
+++ b/src/nvidia-modeset/src/nvkms-framelock.c
@@ -459,7 +459,7 @@ static NVFrameLockEvoPtr AllocFrameLockEvo(NVDevEvoPtr pDevEvo,
         nvEvoLogDev(pDevEvo, EVO_LOG_ERROR, "The firmware on this Quadro Sync "
                     "card is not compatible with the GPUs connected to it."
                     "  Please visit "
-                    "<http://www.nvidia.com/object/quadro-sync.html> "
+                    "<https://www.nvidia.com/object/quadro-sync.html> "
                     "for instructions on installing the correct firmware.");
         goto fail;
     }


Shinyzenith:
@alcaparra did you make this account just to troll? all of your pull requests are spam including this one.

Technetium1 (approved):
@Shinyzenith This is absolutely **_not_** spam. It's a very valid PR with good intent. Not only does it provide a legitimate solution, but it describes what the issue is and why it should be considered a problem. This is the definition of legitimate contribution even if it's a rather insignificant change. Telling or implying that someone is spamming because you don't agree with a contribution is almost definitely very offputting to new contributors.

Shinyzenith:
> @Shinyzenith This is absolutely **_not_** spam. It's a very valid PR with good intent. Not only does it provide a legitimate solution, but it describes what the issue is and why it should be considered a problem. This is the definition of legitimate contribution even if it's a rather insignificant change. Telling or implying that someone is spamming because you don't agree with a contribution is almost definitely very offputting to new contributors.

You do make a compelling point. Although this does make some sense it's a pretty pointless patch. I do apologize if I came off rough to new contributors.

- Aakash

AlexFolland:
It's not pointless though.  It's legit.  The link should use https.

dr460nf1r3 (approved):

Green-Robot-Dev-Studios (approved):

gghhkm:
> You do make a compelling point. Although this does make some sense it's a pretty pointless patch. I do apologize if I came off rough to new contributors.
> 
> -Aakash

Patch? It literally just adds one letter. This isn't a patch as it is just basic security.

ethnh:
@alcaparra Went through and got the rest I could find, See https://github.com/alcaparra/open-gpu-kernel-modules/pull/1

Existential-Kernel:
@alcaparra You should try to replace all the occurrences of http links for the nvidia website. There's a few I found throughout the repository, so I recommend you include the missed ones in your PR.

aritger:
Thanks for the submission; I've merged the specific change in this PR (fixing http://www.nvidia.com/object/quadro-sync.html).  And, I've updated the remaining http => https instances within our internal code base.  That should show up in a subsequent release.

PR #108, opened 2022-05-12
Add basic CI to check build

Davidy22:
Currently only checks that make finishes, but also creates the CI yaml file so that when tests can get added later

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,35 @@
+# .github/workflows/ci.yml
+
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Get tag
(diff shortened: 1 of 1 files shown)


MedzikUser:
#31?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
+# .github/workflows/ci.yml
+
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
Margen67, line 10:
```suggestion
    runs-on: ubuntu-20.04
```

Croydon:
It would be great if you could give a reason when closing pull requests @Binary-Eater 🙂

Binary-Eater:
Fundamentally, this repository is a snapshot of the state of code we publish per driver release. Per release, we have internal CI and QA so maintaining separate CI on the external repository in its current state is redundant. We may revisit this if the model for repository evolves over time.

Issue #110, opened 2022-05-12
trunk/src/common/displayport/src/dp_configcaps.cpp:405: array index sanity check in wrong place ?


dcb314:
Static analyser cppcheck says:

trunk/src/common/displayport/src/dp_configcaps.cpp:405:55: style: Array index 'i' is used before limits check. [arrayIndexThenCheck]

Source code is

                   for (int i = 0; caps.linkRateTable[i] && (i < NV_DPCD_SUPPORTED_LINK_RATES__SIZE); i++)

Suggest move sanity check from after use, to before use.

aaronp24:
Thanks for reporting this. I filed internal bug number 3643714 to track this.

PR #117, opened 2022-05-12
moved array sanity check to before the array access

RealAstolfo:
Fix referring to #110


mmaneetsingh (approved):
Thanks for contributing this change.
Change looks good, taking it in the internal development branch.

mmaneetsingh:
Merged. Thanks for the contribution.

ValZapod:
./src/common/displayport/src/dp_configcaps.cpp	405	err	V781 The value of the 'i' index is checked after it was used. Perhaps there is a mistake in program logic.
From my issue :)

@mmaneetsingh Can you do so that merge requests will be without cover commit? For this you need to be in the commiter field. Also PGP signatures should be stripped, since most of those are fake by github itself.

PR #125, opened 2022-05-12
Comment tone

rhe4n:
While good for a file developed by a single person, the tone this comment sets is not adequate for a community-led project.


Leseratte10:
The original comment is fine, imo. Your change alters the meaning of the comment. The old comment makes it clear that the behaviour is unexpected and not what people would assume. Your version makes it sound like this behaviour is intentional.

Also, this is not "a community-led project". NVIDIA is still leading this project and is just accepting contributions from the community.

peepopoggers:
"watch yo tone"
but yea I agree with the finely polished mii character's comment above me.

mtijanic:
The comment in question is a direct quote from an engineer and should not be modified. The hint is in the quotation marks. Closing.

PR #127, opened 2022-05-12
Migrate to issue forms

TheEvilSkeleton:
GitHub announced [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) a while ago. Issue forms are essentially forms you have to fill out, as opposed to the current Markdown section. It is much easier to use issue forms, looks prettier and some sections may even hard require for user input.

If you want to see it in action, don't hesitate to file issues! https://github.com/TheEvilSkeleton/open-gpu-kernel-modules/issues/new/choose


niv:
Here we go. :) Thanks for the nudge!

PR #142, opened 2022-05-13
Avoid potential compilation error with old gcc

oToToT:
Though it seems that NVOC private feature is not used now, I think
it is still not bad to fix this now in case of any future errors.

The optional message for deprecated attribute is introduced in
gcc 4.5[1]. To avoid potential compilation error, we should check
gcc version before using it.

Also, \_\_GNUG\_\_ is for testing GNU C++ compiler, so we should
remove it as we are targeting C here.

[1] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=43666

### DIFF
--- a/src/nvidia/inc/libraries/nvoc/prelude.h
+++ b/src/nvidia/inc/libraries/nvoc/prelude.h
@@ -243,7 +243,7 @@ typedef struct NVOC_CLASS_INFO
 #pragma warning(error: 1786) // treat deprecated as error (globally affected)
 #define NVOC_PRIVATE_FIELD(x)       __attribute__((deprecated(#x " is a private field"))) x
 #define NVOC_PRIVATE_FUNCTION(x)    __attribute__((deprecated(#x " is a private function"))) x
-#elif defined(__GNUC__) || defined(__GNUG__) // gcc
+#elif __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5) // gcc
 #pragma GCC diagnostic error "-Wdeprecated-declarations" // treat deprecated as error (globally affected)
 #define NVOC_PRIVATE_FIELD(x)       __attribute__((deprecated(#x " is a private field"))) x
 #define NVOC_PRIVATE_FUNCTION(x)    __attribute__((error(#x " is a private function"))) x


Leseratte10:
GCC 4.5 is from 2010, I don't think anyone will ever compile this driver with a compiler that's more than 12 years old...

YusufKhan-gamedev:
> GCC 4.5 is from 2010, I don't think anyone will ever compile this driver with a compiler that's more than 12 years old...

I dont think anyone is running nvc0 cards since thats 12 years old as well...

Leseratte10:
There's a huge difference between running an old GPU (which you may have laying around, and/or is cheaper, and/or because you can't even get a new GPU due to the current pricing situation), and running a 12-year old compiler when you can just download a newer one for free in a couple minutes.

oToToT:
@Leseratte10 @YusufKhan-gamedev I do agree that the use of gcc 4.4 or older is rare, but I believe it is not none.

The 2 main reasons for this patch:
1. Testing `__GNUG__` is here useless and kinda wrong here, so we should remove it.
2. Changing a compiler in some old system is not always that easy as it might break things, so why don't we fix it here? This patch harm nothing (while changing the compiler might) but make the kernel module more correct and fix a real error that users might face. Additionally, if you searching up places like [GitHub Issues](https://github.com/search?q=%22Wrong+number+of+arguments+specified+for+%27deprecated%27+attribute%22&type=Issues), you could still find some people reporting the same problem in recent years.

ValZapod:
> GCC 4.5 is from 2010, I don't think anyone will ever compile this driver with a compiler that's more than 12 years old...

It is still often used by GCC devs:

>Then you can build newer GCC using older GCC. E.g. GCC 4.0.4 -> GCC 4.7.4 -> modern GCC.

From here: https://stackoverflow.com/questions/9429491/how-are-gcc-and-g-bootstrapped/65708958#65708958

Leseratte10:
It's often used by GCC devs _to compile newer GCC versions_, not to compile a kernel module. 
But I agree that the `__GNUG__` check should be removed; and when that code is edited anyways it wouldn't hurt to add that GCC version check too.

Binary-Eater:
Thanks for the contribution. Unfortunately, we cannot allocate resources to QA supporting GCC 4.5, so we will not be able to opt into taking this change.

PR #145, opened 2022-05-13
refactor: remove ternary operator when useless

Helias:
## CHANGES PROPOSED:
-  converted `(X > Y ? true : false)` into `(X > Y)`
-  converted `(X ? true : false)` into `!!X`

I did the same refactoring in a big C++ project here https://github.com/azerothcore/azerothcore-wotlk/pull/3327, so I am sure it works.

### DIFF
--- a/src/common/displayport/src/dp_evoadapter.cpp
+++ b/src/common/displayport/src/dp_evoadapter.cpp
@@ -278,16 +278,16 @@ void EvoMainLink::queryGPUCapability()
         // MST feature on particular sku, whenever requested through INF.
         //
         _hasMultistream         = (params.bIsMultistreamSupported == NV_TRUE) && !_isMstDisabledByRegkey;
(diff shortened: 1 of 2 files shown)


MurdoMaclachlan:
I'm a little confused how the double negative of `!!X` is different from just `X`?

Helias:
> I'm a little confused how the double negative of `!!X` is different from just `X`?

Nice question, if you have an object and you return it, you'll return that object, so sometimes it is important to use something like `object ? true : false`, this means `if this object is not null and it's defined return true, otherwise return false`, so using the `!!` allows you to make this short and convert any variable (object or any) into a boolean.

mtijanic:
Thank you for the PR, but marking as `cosmetic-postponed` for now. Please see https://github.com/NVIDIA/open-gpu-kernel-modules/discussions/190 for details.

Helias:
I totally understand, it's a low priority PR, no worries 👍

PR #150, opened 2022-05-13
Fixed naming convention.

keroeslux:


vkb24312:
Why? This doesn't make the code any clearer. In fact, I would say it makes it harder to read

mtijanic:
Change does not remotely do what the commit message claims. Closing.

PR #155, opened 2022-05-13
[~] Remove deprecated WIN32 defines

TrickyDmitriy:
It's time to remove WIN32 specific names completely.

Also there are generated files 
by **rmconfig** [rmconfig.h] 
and **chip-config** [g_allclasses.h]) that shouldn't be modified manually. 

Unluckily, these generation tools are not available in repo.

### DIFF
--- a/src/common/sdk/nvidia/inc/nvos.h
+++ b/src/common/sdk/nvidia/inc/nvos.h
@@ -355,11 +355,8 @@ typedef struct
 } NVOS05_PARAMETERS;
 
 /* Valid values for hClass in Nv01AllocEvent */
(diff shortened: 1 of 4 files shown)


mtijanic:
Hello and thank you for your interest. Unfortunately, this codebase is shared with various other projects at NVIDIA, including the Windows drivers, so we cannot just remove all references.

In general, a lot of things that seem like noise might be from one of the other projects, so it's always best to check with us before removing anything.

mtijanic:
Additionally, `nvos.h` is part of the API that is used by all usermode clients (CUDA, OpenGL, Vulkan, NVML, etc), so changes to these files would require updating all those projects as well. It's best to perform no cleanup in the `sdk/` folder at all.

PR #164, opened 2022-05-13
~: Fix warnings reported by clang

obiwac:
Fix/silence various warnings reported by clang (version 14.0.1).

### DIFF
--- a/kernel-open/common/inc/nv.h
+++ b/kernel-open/common/inc/nv.h
@@ -180,9 +180,9 @@ extern const char *pNVRM_ID;
 
 typedef union
 {
-    volatile NvV8 Reg008[1];
-    volatile NvV16 Reg016[1];
-    volatile NvV32 Reg032[1];
+    volatile NvV8 Reg008[2564];
+    volatile NvV16 Reg016[1282];
+    volatile NvV32 Reg032[641];
 } nv_hwreg_t, * nv_phwreg_t;
 
 
--- a/src/common/uproc/os/libos-v2.0.0/debug/lines.c
+++ b/src/common/uproc/os/libos-v2.0.0/debug/lines.c
@@ -877,7 +877,7 @@ static void dwarfSetARangeTableLineUnit(libosDebugResolver *pThis, DwarfStream u
     }
 
     // Walk forwards setting the line unit info for all entries with the same arange unit.
-    for (; (foundUnit == pThis->arangeTable[i].arangeUnit); i++)
+    for (; foundUnit == pThis->arangeTable[i].arangeUnit; i++)
     {
         pThis->arangeTable[i].lineUnitBuffer = unit.buffer;
         pThis->arangeTable[i].lineUnitSize   = (NvU32)unit.size;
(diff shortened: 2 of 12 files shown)


ptr1337:
This should also help with the failing of dkms, right?

obiwac:
> This should also help with the failing of dkms, right?

Not sure what exactly you're talking about, but likely not. This only fixes compiler warnings, doesn't a priori fix any bugs in the program itself.

ptr1337:
https://bugs.archlinux.org/task/74714?project=1&string=nvidia-open-dkms

kernel-open/common/inc/nv.h
```diff
@@ -180,9 +180,9 @@ extern const char *pNVRM_ID;
 
 typedef union
 {
-    volatile NvV8 Reg008[1];
-    volatile NvV16 Reg016[1];
-    volatile NvV32 Reg032[1];
+    volatile NvV8 Reg008[2564];
```
imaami, line 183: This doesn't seem right. How is changing the array sizes this drastically an appropriate fix for a compiler warning? Even if the pointer typedef (nv_phwreg_t) is meant to be a convenience type for register address casts how can you be sure that the non-pointer type (nv_hwreg_t) isn't used by some external codebase in a way that depends on the size? And is the new size guaranteed to be 2564 bytes everywhere and always? If it is then why use a magic number instead of a macro with an informative name?

obiwac: > how can you be sure that the non-pointer type (nv_hwreg_t) isn't used by some external codebase in a way that depends on the size?

I highly doubt it's anything more than a convenience type. It wouldn't make sense for each field to have a length of 1 otherwise.

> And is the new size guaranteed to be 2564 bytes everywhere and always - and even if it is, why use a magic number instead of a macro with an informative name?

That's very fair, I would've hoped a potential reviewer could show a better way of doing things ;) Either way, I don't think this is worse than leaving a 1 and cluttering output with warnings which can easily be silenced.

mtijanic:
Thank you for the PR. I'm postponing this for now and marking as cosmetic, per #190 . While it does provide an immediate benefit of resolving the warnings, we can't actually merge these changes due to shared codebase constraints.

By the way, usually the suggested way of dealing with unused variables in the codebase is `PORT_UNREFERENCED_VARIABLE(var);` rather than deleting them. That should silence the warnings and not break any other build using the same code.

niv:
Thank you for your interest in contributing to open-gpu-kernel-modules.

As recently explained in [CONTRIBUTING.md](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/main/CONTRIBUTING.md), we have decided to not accept cosmetic changes for now.

All changes going into this repository incur significant overhead on our end, and we'd like to focus our resources on executable code changes; so we have decided to close all PRs that make no code changes.

If and when this position changes, your PR will still be in the archives and revisited (as long as it's still tagged cosmetic-postponed).

Please do not feel discouraged to submit other changes! In addition, we will be happy to keep the conversation going on the Discussion boards here on GitHub, and elsewhere.

PR #168, opened 2022-05-14
refactor: remove useless conditions

drendog:
A refactor to remove useless conditions

### DIFF
--- a/kernel-open/nvidia-drm/nvidia-drm-crtc.c
+++ b/kernel-open/nvidia-drm/nvidia-drm-crtc.c
@@ -509,9 +509,9 @@ static int nv_drm_plane_atomic_set_property(
         nv_drm_plane_state->fd_user_ptr = u64_to_user_ptr(val);
 #endif
         return 0;
-    } else {
-        return -EINVAL;
     }
+
+    return -EINVAL;
 }
 
 static int nv_drm_plane_atomic_get_property(
@@ -524,9 +524,9 @@ static int nv_drm_plane_atomic_get_property(
 
     if (property == nv_dev->nv_out_fence_property) {
         return 0;
-    } else {
-        return -EINVAL;
     }
+    
+    return -EINVAL;
 }
 
 static struct drm_plane_state *
--- a/kernel-open/nvidia-uvm/uvm_ampere_ce.c
+++ b/kernel-open/nvidia-uvm/uvm_ampere_ce.c
@@ -70,11 +70,10 @@ static NvU32 ce_aperture(uvm_aperture_t aperture)
     else if (aperture == UVM_APERTURE_VID) {
         return HWCONST(C6B5, SET_SRC_PHYS_MODE, TARGET, LOCAL_FB);
     }
-    else {
-        return HWCONST(C6B5, SET_SRC_PHYS_MODE, TARGET, PEERMEM) |
-               HWVALUE(C6B5, SET_SRC_PHYS_MODE, FLA, 0) |
-               HWVALUE(C6B5, SET_SRC_PHYS_MODE, PEER_ID, UVM_APERTURE_PEER_ID(aperture));
-    }
+    
+    return HWCONST(C6B5, SET_SRC_PHYS_MODE, TARGET, PEERMEM) |
+           HWVALUE(C6B5, SET_SRC_PHYS_MODE, FLA, 0) |
+           HWVALUE(C6B5, SET_SRC_PHYS_MODE, PEER_ID, UVM_APERTURE_PEER_ID(aperture));
 }
 
 // Push SET_{SRC,DST}_PHYS mode if needed and return LAUNCH_DMA_{SRC,DST}_TYPE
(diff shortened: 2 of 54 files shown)


grtcdr:
> A refactor to remove useless conditions

You have not done that though.

```c
if (cond) {
  return X;
} else {
  return Y;
}
```

Is no different, performance-wise from:

```c
if (cond) {
  return X;
}

return Y;
```

One is just more readable than the other.

sTiKyt:
Imothis refactor just makes readability worse

grtcdr:
> Imo this refactor just makes readability worse

I agree. Also, the fact that the developers are very consistent with the use of `else` means that there's a reason for its existence.

mtijanic:
Thank you for the PR, but marking as `cosmetic-postponed` for now. Please see https://github.com/NVIDIA/open-gpu-kernel-modules/discussions/190 for details.

cjwijtmans:
> > A refactor to remove useless conditions
> 
> You have not done that though.
> 
> ```c
> if (cond) {
>   return X;
> } else {
>   return Y;
> }
> ```
> 
> Is no different, performance-wise from:
> 
> ```c
> if (cond) {
>   return X;
> }
> 
> return Y;
> ```
> 
> One is just more readable than the other.

I disagree in this case. You expect a function to end with a return and not an else return. The latter makes no sense in flow even though its completely fine to compile and is the same.

Alcaro:
In my opinion, if one path through the function is normal and the others are errors, the normal path's return should be at the top level. For example, [src/common/modeset/timing/nvt_edid.c](https://github.com/NVIDIA/open-gpu-kernel-modules/pull/168/files#diff-241432f433b3790843e3d5d90990f66bef49143cf3dd4ae971520415eb3437f7) and [src/nvidia/src/lib/protobuf/prbutil.c](https://github.com/NVIDIA/open-gpu-kernel-modules/pull/168/files#diff-463e22ede67962100bda1da57200a86e6fd5badb9b8e17ff76aab476f12ca261) look like improvements to me.

If each path is equally normal, like [kernel-open/nvidia-uvm/uvm_ampere_mmu.c](https://github.com/NVIDIA/open-gpu-kernel-modules/pull/168/files#diff-3b0c84bc161792ec8d48661aaaab8d6262a18a1cc43a06a130f0dcddd8e8634e), it looks better to keep each return at the same indentation level; I'd decline that change.

But that's just my opinion, and I don't know how common it is. The nice thing about code styles is that you have so many to choose from.

grtcdr:
@cjwijtmans 

> *[...]* The latter makes no sense in flow **even though its completely fine to compile and is the same.**

I disagree that it makes no sense. For the compiler it absolutely does, like you mentioned. This is simply a matter of preference.

cjwijtmans:
> @cjwijtmans
> 
> > _[...]_ The latter makes no sense in flow **even though its completely fine to compile and is the same.**
> 
> I disagree that it makes no sense. For the compiler it absolutely does, like you mentioned. This is simply a matter of preference.

Yeah but the main feature of a human readable language is that is readable by a human not a compiler. A function always returns something unless void I know this and the compiler knows this. A conditional return is not very readable unless the condition is in the return statement itself. Thus you dont else return, you always return. It could be a preference but to me for human readability and coding flow else return dont make much sense to me even though in the backend the result is the same. But you might as well do a lot of other shady things because the compiler deals with it anyway, that seems like a fallacy to me.

grtcdr:
I would love to discuss this further with you, but that wouldn't really change anything as this pull request has been labeled "cosmetic-postponed" and I highly doubt it'll ever be accepted, it doesn't do what it states it does. The fact this label exists is due to the excessively large amount of "useless" (IMO) PR that this project has been receiving lately.

cjwijtmans:
Its not useless though. Quality of code even cosmetic will make the project easier to maintain and in extreme cases even prevent bugs.

shkhln:
> Its not useless though. Quality of code even cosmetic will make the project easier to maintain and in extreme cases even prevent bugs.

Keep in mind that this driver code exists in multiple repos and branches. Large cosmetic PRs would have to be adapted to all these versions (which is tedious and error-prone) otherwise the difference would it make more difficult to backport bugfixes and whatnot. This is a fair amount of work for very little benefit.

niv:
Thank you for your interest in contributing to open-gpu-kernel-modules.

As recently explained in [CONTRIBUTING.md](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/main/CONTRIBUTING.md), we have decided to not accept cosmetic changes for now.

All changes going into this repository incur significant overhead on our end, and we'd like to focus our resources on executable code changes; so we have decided to close all PRs that make no code changes.

If and when this position changes, your PR will still be in the archives and revisited (as long as it's still tagged cosmetic-postponed).

Please do not feel discouraged to submit other changes! In addition, we will be happy to keep the conversation going on the Discussion boards here on GitHub, and elsewhere.

SingleSlice:
L

PR #191, opened 2022-05-14
uapi: Use proposed libdrm uapi header

YusufKhan-gamedev:
Changes to make the MR from libdrm work better without implementing ugly userland hacks


YusufKhan-improves-shit:
Note: This pretty much just allows C++ files to include this(in the context of this driver).

Issue #102, opened 2022-05-12
PVS-Studio big list of warnings


ValZapod:
[projectcrazy.txt](https://github.com/NVIDIA/open-gpu-kernel-modules/files/8678843/projectcrazy.txt)

For example, 

```
./src/common/nvswitch/kernel/lr10/lr10.c	5957	warn	V519 The 'status' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 5952, 5957.
./src/common/nvswitch/kernel/lr10/soe_lr10.c	2191	warn	V519 The 'pHal->getEmemSize' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 2189, 2191.
./src/nvidia-modeset/src/nvkms-evo.c	2936	warn	V519 The 'appliedHeadMask' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 2927, 2936.
./src/nvidia/arch/nvalloc/unix/src/osapi.c	4867	note	V519 The 'fbSize' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 4856, 4867.
./src/nvidia/src/kernel/gpu/bus/arch/ampere/kern_bus_ga100.c	417	warn	V519 The 'status' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 411, 417.
./src/nvidia/src/kernel/gpu/ce/arch/ampere/kernel_ce_ga100.c	526	warn	V519 The 'status' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 523, 526.
./src/nvidia/src/kernel/gpu/fifo/arch/volta/usermode_api_gv100.c	60	note	V519 The 'status' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 55, 60.
./src/nvidia/src/kernel/gpu/mem_mgr/arch/maxwell/mem_utils_gm107.c	114	warn	V519 The 'memAllocParams.attr' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 113, 114.
./src/nvidia/src/kernel/gpu/mem_mgr/arch/maxwell/mem_utils_gm107.c	797	warn	V519 The 'rmStatus' variable is assigned values twice successively. Perhaps this is a mistake. Check lines: 789, 797.
[...]

PR #195, opened 2022-05-15
nv-pci: fixed always true expression

ValZapod:


mtijanic (approved):
Great find, thank you for the patch. Pulling it into our internal development branch in p4.

Will update here when we know which releases will pick it up.

ValZapod:
> Great find, thank

This is from my PVS-Studio issue. #102

PAR2020:
Fixed in a future release - 31351203.

Issue #55, opened 2022-05-12
Copy paste bug in `osapi.c`


qarmin:
**NVIDIA Driver Version**
Git 

https://github.com/NVIDIA/open-gpu-kernel-modules/blob/69791e9bf5161c786d620312fd6396756c37cff9/src/nvidia/arch/nvalloc/unix/src/osapi.c#L4589

`pNumaMemAddr` is checked twice if is equal to null

Benedani:
Is it supposed to be pNumaMemSize? That's the one that isn't checked for null.

PAR2020:
Internal bug 3643074

PAR2020:
Resolved in PR #61 initially.

PR #61, opened 2022-05-12
Collection of minor fixes to issues

Joshua-Ashton:

PR #201, opened 2022-05-15
Add NULL check for pNumaMemAddr

fares-soliman:
pNumaMemSize is not NULL checked (instead, pNumaMemAddr is checked twice) 
Fixes: #55


PAR2020:
Dup of #61

ViniciusFXavier (commented):
👍

PR #203, opened 2022-05-15
Added very slight optimisations and general code changes

Existential-Kernel:
Here's a summary of my changes from (arguably) most to least useful in  `src/common/displayport/src/dp_configcaps.cpp`:
- I inlined quite a number of functions that are either short or simple in terms of their operations
- I converted a few if else chains into switch statements, which would fit more accurately than multiple if else conditions
- I modified a few parts of the code so some values are not initialised as undefined, and some parts where the logic can be optimised
- Specified `const` in unchanged parameters
- I also added a few changes in how the whitespace is implemented throughout the code, since most of them are either unnecessary or has no meaningful reason why it's added. This is arguably a pointless change so please let me know if this should remain or be removed in this PR.


monyarm:
I wouldn't call this cosmetic, the whitespace changes sure. But the rest is very real optimizations.

mtijanic:
Hi, @monyarm , thank you for the PR. I have tagged it this way because it is not merge-able for the same reasons cosmetic changes are not mergeable, which is that it does not conform to our styles and guidelines. You couldn't have known this, because we didn't actually publish those, so I really apologize for the churn.

We will be revisiting these PRs after the guidelines are published.

By the way,
> But the rest is very real optimizations.

Even with `-O0` GCC will inline all those functions regardless of the `inline` keyword. And adding `const` to integer-y parameters passed by value does not have a runtime effect - all it does is allow for _compile_ time checks.

carxt:
For what is worth, the inline keyword won't do anything with virtual functions, as those are resolved in runtime, so the compiler cannot remove the call.

PR #204, opened 2022-05-16
Remove outdated macros

jserv:
Macro WARN() is defined in Linux header include/asm-generic/bug.h, and
it is not necessary to preserve the outdated defintions since the NVIDIA
Open GPU kernel module supports Linux v3.10+.

### DIFF
--- a/kernel-open/nvidia-modeset/nv-kthread-q.c
+++ b/kernel-open/nvidia-modeset/nv-kthread-q.c
@@ -48,17 +48,6 @@
 // into the queue, and those functions will be run in the context of the
 // queue's kthread.
 
-#ifndef WARN
-    // Only *really* old kernels (2.6.9) end up here. Just use a simple printk
-    // to implement this, because such kernels won't be supported much longer.
-    #define WARN(condition, format...) ({                    \
-        int __ret_warn_on = !!(condition);                   \
-        if (unlikely(__ret_warn_on))                         \
-            printk(KERN_ERR format);                         \
-        unlikely(__ret_warn_on);                             \
-    })
-#endif
-
 #define NVQ_WARN(fmt, ...)                                   \
     do {                                                     \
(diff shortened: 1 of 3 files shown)


niv:
Thank you for your contribution! Appreciate it. Looks safe to remove. However, we'll get back to it once we've ironed out the code/style guide.

YusufKhan-gamedev:
:( 2.6.9 will be missed

gghhkm:
> :( 2.6.9 will be missed

2.6.9 going EOL just shows how much progress we've made in the newer kernels.

Ruzie:
> :( 2.6.9 will be missed

Today's most of Linux distros doesn't uses 2.6.9 or lower version of kernel, so far it looks better to remove. Except if you're not using DSL.

gghhkm:
@YusufKhan-gamedev 
Why would anyone want to use such an old kernel anyway? What about it keeps people coming back? It came out in 2005, it's 2022 now.

YusufKhan-gamedev:
@gghhkm 
Though *most* people use upstream kernels, there are a *lot* of devices that require old or outdated kernels since they require ugly hacks to the older kernel and even uglier hacks to the modern kernels. Only really a couple devices work on mainline in my house(ie. my laptops and my desktops).

Though I doubt 99.9% of people will be using a RTX 2060 on their old phone, it is still possible and may be needed for them.

gghhkm:
> @gghhkm Though _most_ people use upstream kernels, there are a _lot_ of devices that require old or outdated kernels since they require ugly hacks to the older kernel and even uglier hacks to the modern kernels. Only really a couple devices work on mainline in my house(ie. my laptops and my desktops).
> 
> Though I doubt 99.9% of people will be using a RTX 2060 on their old phone, it is still possible and may be needed for them.

meanwhile... me... if it can't even run rust-lang I chuck it in the trash lmao.

Issue #18, opened 2022-05-11
No documentation on which GPUs this driver supports


RobertCochran:
There is no bleedingly obvious list of what GPU models this driver does and does not support. It's of course safe to assume that this driver supports all of the latest GPUs and at least a generation back, but beyond that it's difficult to ascertain whether you're in the green here. I know that the proprietary driver fairly regularly drops support for older cards. What options are available for folks that have said older cards?

Please provide a clear list of GPUs are that supported by this driver, thanks.

Aang23:
It's very 515, so it will support what the equivalent proprietary driver supports. Architecture-wise, that includes : 
Maxwell, Pascal, Turing, Ampere, Volta.

Personally though, I wouldn't mind if the code for Kepler could be released as well.

Binary-Eater:
My personal understanding is that it essentially supports any card that is able to use the GSP firmware. That information seems to be available in the documentation provided with each NVIDIA driver release.

https://download.nvidia.com/XFree86/Linux-x86_64/515.43.04/README/gsp.html

sempervictus:
Based on the source code here and in the last binary driver set, it looks to "just be the next version" of the same driver, so the same chipsets.

9p4:
This is limited to only GSP-enabled cards, so 20xx and 30xx cards (Turing and above).

From https://www.phoronix.com/scan.php?page=article&item=nvidia-open-kernel&num=1

> Well, it wouldn't entirely since let's not forget this new kernel driver only works with Turing GPUs and newer. Meanwhile the older sweet spot currently for the Nouveau kernel driver is GTX 600/700 Kepler series (and GTX 750 Maxwell1) but the newer GPUs is where it lacks proper power management firmware / re-clocking and thus the performance is in bad shape. Nouveau on recent generations of NVIDIA GPUs is basically only good enough for driving a display as due to the ability to re-clock to the peak performance state its 3D graphics performance is frankly junk. So the NVIDIA Open Kernel Driver is certainly superior for GeForce RTX 20/30 series while GTX 900 / GTX 10 graphics cards will likely be left in an awkward state outside of the proprietary driver stack.

Aang23:
> This is limited to only GSP-enabled cards, so 20xx and 30xx cards (Turing and above).
> 
> From https://www.phoronix.com/scan.php?page=article&item=nvidia-open-kernel&num=1
> 
[...]

PR #207, opened 2022-05-16
Added Supported GPUs to README

balanarayank:
Added section for Supported GPU models. This can be updated going forward. In future, it can be considered to reference a link to a new Nvidia website page that would provide a list of supported GPU models for open kernel modules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,13 @@ E.g.,
         OBJCOPY=aarch64-linux-gnu-objcopy
 
 
+## Supported GPU models
+
+- NVIDIA Ampere and Turing architecture GPUs are supported.
+- Datacenter GPUs are supported for production.
+- GeForce and Workstation GPUs support is currently alpha quality.
+
+
 ## Other Build Knobs
 
 NV_VERBOSE - Set this to "1" to print each complete command executed;
--- a/README.md
+++ b/README.md
@@ -50,8 +50,42 @@ E.g.,
 
 ## Supported GPU models
 
-- NVIDIA Ampere and Turing architecture GPUs are supported.
-- Datacenter GPUs are supported for production.
+- <details>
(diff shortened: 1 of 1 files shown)


LinusCDE:
> I agree, but the README would become too large if all the supported GPU models are mentioned here.

I don't see the problem in length when just mentioning multiple generations to clarify where a potential support cut-off would be. There is no need to mention each and every model explicitly imo.

> In my previous message, my intention was to suggest that it's better to have a web page created on Nvidia website/forums that lists all the supported GPU models with their actual names [...]

That seems a nice idea. Does nvidia already have such a page? If not it might be too much to ask then. But I like the idea.

balanarayank:
> That seems a nice idea. Does nvidia already have such a page? If not it might be too much to ask then. But I like the idea.

I think such a page would definitely help regular users going forward. They can easily refer to see if their GPU is supported or not. Perhaps someone from Nvidia side would be able to provide more insight into this.

gghhkm:
> I agree, but the README would become too large if all the supported GPU models are mentioned here. 

<details><summary>Do This</summary>
<Architecture Name> (GTX 7xx to GTX/RTX 16xx series) 
</details>

```
<details><summary>Do this</summary>
 Text
</details>
```

mtijanic:
Thank you everyone for the ideas and examples. We are aware this is an important bit we missed.

These lists actually have to be generated from our internal databases and updated automatically as products are released, which is not something the community can implement. I've tagged it with NV-Triaged.

aritger:
Yes, thanks for the ideas.  We're going to provide a list of compatible GPUs, including PCI IDs, in a future release.

This PR overlaps quite a bit with https://github.com/NVIDIA/open-gpu-kernel-modules/issues/18 I'm going to close this PR, and a future release will close that Issue.

Issue #109, opened 2022-05-12
Null pointer dereference in nv-acpi.c


tallossos:
**NVIDIA Driver Version**
Git

https://github.com/NVIDIA/open-gpu-kernel-modules/blob/8d7f201155cc257bbef784a7887a6ff737f2fb5d/kernel-open/nvidia/nv-acpi.c#L718-L720
https://github.com/NVIDIA/open-gpu-kernel-modules/blob/8d7f201155cc257bbef784a7887a6ff737f2fb5d/kernel-open/nvidia/nv-acpi.c#L737-L746 There isn't any validation on the return value from `acpi_bus_get_device` unlike other places where  `acpi_bus_get_device` is being used.
`device` might be null when dereferencing.

amada95:
I've addressed this issue in [this pull request](https://github.com/NVIDIA/open-gpu-kernel-modules/pull/22)

PAR2020:
Since PRs are closed, will close this issue as well.

PR #216, opened 2022-05-17
Fix device null pointer deref and add acpi bus query failure check

amada95:
Makes use of `ACPI_FAILURE` macro to check successful operation of `acpi_bus_get_device`, and adds a null check before using the contents of `device`.

Fixes: #109 

_note: functional change, followup from PR #22_

### DIFF
--- a/kernel-open/nvidia/nv-acpi.c
+++ b/kernel-open/nvidia/nv-acpi.c
@@ -717,6 +717,7 @@ acpi_status nv_acpi_find_methods(
 
 void NV_API_CALL nv_acpi_methods_uninit(void)
 {
+    acpi_status status;
     struct acpi_device *device = NULL;
 
     nvif_handle = NULL;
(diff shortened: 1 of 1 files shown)


jarodsim:
Yes, i think this fix work great

mtijanic (changes requested):
Hi @amada95 , thank you for the contribution.

I see a couple of issues that need to be worked out still:
1. The newly added `status` variable is not declared anywhere. Please declare it at the top in line with the style of this file. Additionally, please at least ensure your changes compile successfully. 
2. The check is done in an `#if defined(NV_ACPI_BUS_GET_DEVICE_PRESENT)`. Should this be undefined, the check will be skipped and a null deref remains at `device->driver_data = NULL;`
3. In case of function call failure, we return early, but I believe we would still want to set `nvif_parent_gpu_handle` to NULL.

amada95:
The requested changes have been implemented, let me know if there are any issues.

mtijanic:
Thank you, Amada, this looks good now. I'll start the internal test pipeline and get back to you on Monday.

master7720:
i use arch btw.

amada95:
Hello! Just popping in, any updates on this PR?

mtijanic:
Hi Amada, sorry for the late response, I was out of office for a few days.

The status is this: This code has been removed from our primary development branch. It is still around in some of the release branches, where I expect we will merge it. However, for the very next release(s?), QA has already advanced too far and the branch is currently frozen. This means that if I merge the PR here, it would be clobbered by releases that are already in the pipeline.

So, I'll leave it open for the time being and merge when it gets merged internally.

Thanks for understanding.

amada95:
Sounds good, thanks for the transparency and I wish you all the best of luck with that.

amada95:
Accidentally renamed the main branch, reopening

amada95:
Hello! Any updates? I just want to make sure this didn't get buried due to that accidental branch rename.

mtijanic:
Hey there, sorry for the late response.  From the previous comment:

> The status is this: This code has been removed from our primary development branch. It is still around in some of the release branches, where I expect we will merge it.

Looks like the function removal actually did make it into the next release QA pipeline, so the entire function is already gone from the repo (which is why github is telling you there are merge conflicts).

I don't think there's anything left to do here, so I'll close the PR. Please let me know if you disagree. And thanks again for the contribution, even if it didn't get merged!

PR #224, opened 2022-05-18
Implement CRTC_X/Y for primary planes

Joshua-Ashton:
This is used by Gamescope for centring the game's buffer.

Adds a check to see if the `supportsWindowMode` cap is supported
for the main plane (EVO3+) and implements CRTC_X/Y for
NVKMS_MAIN_LAYER if it is supported.

This was not checked before, so CRTC_X could be specified on the main
layer, and the atomic commit would succeed, resulting in incorrect
output.


liam-middlebrook:
Thanks for contributing this change @Joshua-Ashton, either I or someone else from NVIDIA will get back to this change soon. In the meantime I've filed NVIDIA-internal bug 3649215 to track this contribution.

src/nvidia-modeset/kapi/src/nvkms-kapi.c
```diff
@@ -2432,6 +2435,18 @@ static NvBool NvKmsKapiPrimaryLayerConfigToKms(
         changed = TRUE;
     }
 
+    if (layerRequestedConfig->flags.dstXYChanged || bFromKmsSetMode) {
```
liam-middlebrook, line 2438: nit: Could you please move this after the `srcXYChanged` block below?

src/nvidia-modeset/kapi/src/nvkms-kapi.c
```diff
@@ -2432,6 +2435,18 @@ static NvBool NvKmsKapiPrimaryLayerConfigToKms(
...
+        /* If the main layer doesn't support window mode,
```
liam-middlebrook, line 2439: nit: Please align the `*`'s in block comments and have the comment close on a newline like so:
```
/*
 * If the main layer...
 * ...
 */
```

liam-middlebrook (changes requested):

src/nvidia-modeset/kapi/src/nvkms-kapi.c
```diff
@@ -2432,6 +2435,18 @@ static NvBool NvKmsKapiPrimaryLayerConfigToKms(
...
+            * then return false here to fail the commit/test. */
+        if (!device->caps.mainLayerSupportsWindowMode &&
+                (layerConfig->dstX != 0 || layerConfig->dstY != 0))
```
liam-middlebrook, line 2442: nit: Please don't leave floating conditional blocks.

liam-middlebrook:
I applied this patch locally and when testing it out. noticed some incorrect behavior after the gamescope process exits.

On a system with an RTX 2070, if I run `./gamescope -- glxgears -geometry 1800x1080+0+0`, I notice that this patch does indeed center the gamescope layer on the display, but after exit it does not reset the layer position properly, such that my console is offset by `60px` to the right. If I unload the nvidia-drm kernel module, the layer position resets to `0,0` as expected.

When you tested this did it work for you, or is there a similar issue where the position is "stuck" until nvidia-drm is unloaded?

I did a bit of cursory debugging on this and found that upon closing of gamescope on my system the `NVC57B_SET_POINT_OUT` method is never getting sent to the display engine, so the console is not restored to `0,0`.

Joshua-Ashton:
@liam-middlebrook Will get back to this when I am at PC with NV GPU again (next week).

PAR2020:
@Joshua-Ashton, still waiting on your comments so the change can be checked in and released...
Thanks!

PR #227, opened 2022-05-18
always clear numPagesAlloc

trixirt:
Clang static analysis on RHEL reports this issue
addrtree.c:1154:50: warning: The left operand of '/' is a garbage value [core.UndefinedBinaryOperatorResult]
  *numPagesAlloc += localNumPagesAlloc / 2;
                    ~~~~~~~~~~~~~~~~~~ ^

_pmaAddrtreeScanContiguous() can return early with an error without
setting numPagesAlloc.  There are several uses of
_pmaAddrtreeScanCantiguous() that do not check the status and
depend on the setting of of numPageAlloc as a side effect.

To ensure the side effect is always valid, move the clearing
of numPagesAlloc to before the early exit

Signed-off-by: Tom Rix <trix@redhat.com>

### DIFF
--- a/src/nvidia/src/kernel/gpu/mem_mgr/phys_mem_allocator/addrtree.c
+++ b/src/nvidia/src/kernel/gpu/mem_mgr/phys_mem_allocator/addrtree.c
@@ -848,6 +848,8 @@ _pmaAddrtreeScanContiguous
     // This requirement is ensured in PMA
     NV_ASSERT(alignment >= pageSize && portUtilIsPowerOfTwo(alignment));
 
+    *numPagesAlloc = 0;
+
     // Only focus on the level above the pageSize level. Children are ignored.
     level = addrtreeGetTreeLevel(pageSize);
     if (level == 0)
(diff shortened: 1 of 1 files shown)


niv:
Hi! Thanks for checking back. I was out for a bit, but I'll handle it soon. This hasn't been forgotten.

trixirt:
PR has stalled

YusufKhan-gamedev:
Could the real issue be that `numPagesAlloc` is being modified even when `_pmaAddrtreeScanContiguous` fails?
```

            status = _pmaAddrtreeScanContiguous(
                pMap,
                addrBase,
                rangeStart,
                rangeEnd,
                2,
                freeList + i,
                _PMA_64KB,
                alignment,
                &localNumPagesAlloc,
                bSkipEvict);
            // Give back the first of every two 64KB frames
            *numPagesAlloc += localNumPagesAlloc / 2;
            if (status != NV_OK)
            {
                return status;
            }
```

PR #254, opened 2022-05-24
Fix null pointer derefs in src/nvidia-modeset and src/common/nvlink

amada95:
Added null checks to avoid undefined behaviour caused by potential null pointer dereferences at runtime.

### DIFF
--- a/src/common/nvlink/kernel/nvlink/core/nvlink_conn_mgmt.c
+++ b/src/common/nvlink/kernel/nvlink/core/nvlink_conn_mgmt.c
@@ -44,7 +44,7 @@ nvlink_core_get_intranode_conn
 
     FOR_EACH_CONNECTION(tmpConn, nvlinkLibCtx.nv_intraconn_head, node)
     {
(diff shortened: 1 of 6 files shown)


aaronp24:
I don't think these list iterator variables can ever be NULL unless there's memory corruption due to the way the macro is defined:
```
#define nvListForEachEntry(pos, head, member)                           \
    for (__NV_LIST_SET(pos, __nv_container_of((head)->next, pos, member));            \
         &pos->member != (head);                                        \
         __NV_LIST_SET(pos, __nv_container_of(pos->member.next, pos, member)))
```
For `pos` to become NULL, `(head)->next` would need to be NULL (which it should never be), _and_ `member` would need to be the first member of the containing struct so that the `__nv_container_of((head)->next, pos, member)` expression also returns NULL.

amada95:
I see, that makes sense; in addition, with that in mind, the change in  `kern_perf_gpuboostsync.c` would also be unnecessary, as the `GPUBOOSTMGR_ITR_START` macro seems to have all necessary null checks baked-in already. Thanks for the feedback, I'm going to go ahead and close this PR.

PR #267, opened 2022-05-27
Update nvlimits.h for readability.

blockhead100:
Remove extra lines


aaronp24:
Thanks for the report. Extra whitespace like this will be handled in the future by changes to our internal tooling. See for example #93 and #79.

PR #273, opened 2022-05-31
` is deprecated

arch-user-france1:
` is deprecated (ref: https://wiki.bash-hackers.org/scripting/obsolete and the script called shellcheck). I replaced them with $().

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,12 +8,12 @@ version 515.43.04.
 
 To build:
 
-    make modules -j`nproc`
+    make modules -j$(nproc)
 
 To install, first uninstall any existing NVIDIA kernel modules.  Then,
 as root:
 
-    make modules_install -j`nproc`
+    make modules_install -j $(nproc)
 
 Note that the kernel modules built here must be used with gsp.bin
 firmware and user-space NVIDIA GPU driver components from a corresponding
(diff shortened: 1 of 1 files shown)


arch-user-france1:
> My shell program removed this a long time ago. `Acked-by: Yusuf Khan <yusisamerican AT gmail DOT com>`

Note that it just is not recommended to use but it will stay there. No plans for removing it.

arch-user-france1:
Okay, well then, I will close it.

aritger:
@MilesBHuff: we had actually already pulled this change into our code base, and it is now included in the 515.57 release.  Do you have an example shell that doesn't honor the `$()` syntax?

aritger:
@arch-user-france1: note that we already applied this change to our tree, and it is present in 515.57.  I don't think there is anything else to be done in this pull request unless someone can point to a shell that doesn't support the `$(nproc)` syntax, and then we would need to revert the change.

I'm going to go ahead and re-close this.  Let me know if I've misunderstood.

MilesBHuff:
@aritger 
> Do you have an example shell that doesn't honor the $() syntax?

The original Bourne shell.

MilesBHuff:
@BratishkaErik
>  https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03

Interesting!  I stand corrected, then:  I guess it *is* POSIX-compliant.

birdie-github:
It's been deprecated for years if not a decade, it's not gonna get removed now or ever. I wouldn't say this merge request is useless but there's a scent of it.

PR #280, opened 2022-06-03
Return EINVAL when allocating memory and support NULL padding

YusufKhan-gamedev:
Signed-off-by: Yusuf Khan <yusisamerican AT gmail DOT com>


aaronp24:
`0` and `NULL` have the same numerical value so these extra checks are redundant. These `__pad` values have `uint` types so comparing them against 0 is appropriate.

In addition, using `|` for logical or is unusual.

PR #296, opened 2022-06-07
Add Objective-C class to nvidia-drm

ColinTheContributor:
I heard from some guy that Objective-C is better than C for writing drivers. So I added an objective C class because it makes the driver better .


aritger:
Thanks, but I think we'll stick with C for now.

PR #308, opened 2022-06-16
Fix/Improve Makefile

izenynn:
fix: add rule "all" to PHONY.
chore: improve PHONY declaration for readability.
refactor: split variables declaration area and rules area.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,9 +6,11 @@
 # To install the build kernel modules: run (as root) `make modules_install`
 ###########################################################################
 
-include utils.mk
+###########################################################################
+# variables
+###########################################################################
 
-all: modules
+include utils.mk
 
 nv_kernel_o                = src/nvidia/$(OUTPUTDIR)/nv-kernel.o
 nv_kernel_o_binary         = kernel-open/nvidia/nv-kernel.o_binary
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -6,23 +6,30 @@
 # To install the build kernel modules: run (as root) `make modules_install`
 ###########################################################################
 
-include utils.mk
```
izenynn, line 9: |   Added line again below
\/

aaronp24:
Adding `all` to `.PHONY` is a good fix, but I don't feel like the rest of the change is an improvement. Adding rules to a new `$(PHONY)` and then declaring `.PHONY: $(PHONY)` at the end just adds clutter.

izenynn:
well that's true for short makefiles, but for large ones things are different, that type of declaring phony make it easier to add/remove entries and it's more readable that having all the rules together in a phony (imagine if you need to scroll down to phony to remove a rule, and then add it, maybe i'm way too much laziness).

Makefile
```diff
@@ -6,23 +6,30 @@
...
+###########################################################################
+# variables
+###########################################################################
 
-all: modules
+include utils.mk
```
aritger, line 14: Inclusion of utils.mk isn't really a variable.  Would it make sense to move this include above the `variables` comment?

aritger:
The Makefile already does `.PHONY: something` multiple times, so I'm pretty sure that is safe to do.  If you want to reorder, let's just do the
```
all: a b
.PHONY: all

a:
	true
.PHONY: a

b:
	true
.PHONY: b
```
pattern you describe, instead of using an intermediate `$(PHONY)` variable.

Makefile
izenynn: good point, let me fix it

izenynn:
I commited the changes you suggested

PR #328, opened 2022-07-20
DRAFT: uvm push uapi

YusufKhan-gamedev:
This is an _experimental_ uapi for exposing the pushbuffer to userspace through ioctl. This needs to have error checking/pushbuf validation, and support for pushes with a size greater than 6 before it can be merged.

Im not going to _really_ do much with this until I need to use it.


aaronp24:
Unless we have a specific use case for something like this, let's just close this for now.

PR #330, opened 2022-07-22
Introduce NV_KZALLOC macro

trixirt:
The pattern of
  p = kmalloc(...);
  memset(p, 0, ...);

Can be reduced to
  p = kzalloc(...)

So introduce and use NV_KZALLOC, a macro that wraps kzalloc.

Signed-off-by: Tom Rix <trix@redhat>

### DIFF
--- a/kernel-open/common/inc/nv-linux.h
+++ b/kernel-open/common/inc/nv-linux.h
@@ -605,6 +605,13 @@ static NvBool nv_numa_node_has_memory(int node_id)
             NV_MEMDBG_ADD(ptr, size); \
     }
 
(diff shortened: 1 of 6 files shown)


mtijanic:
Thanks Tom, I've merged this into our internal development branch. Due to complex release schedules, it's easiest to just let it naturally bubble up in one of the next major releases (52x.xx)

This means that if I merge the PR now into `main` on github, the changes would be clobbered with the next 515.xx point release comes. So I'm tagging as implemented, and we'll merge once we know the very next release will pick it up.

Please let me know if there's a particular need to backport into 515. And thanks again for the contribution!

PR #355, opened 2022-08-24
drm/prime: support dmabuf mmap

YusufKhan-gamedev:
Not sure if its 'prime' or 'PRIME'

Signed-off-by: Yusuf Khan <yusisamerican@gmail.com>

### DIFF
--- a/kernel-open/nvidia-drm/nvidia-drm-drv.c
+++ b/kernel-open/nvidia-drm/nvidia-drm-drv.c
@@ -826,7 +826,9 @@ static struct drm_driver nv_drm_driver = {
     .gem_prime_get_sg_table = nv_drm_gem_prime_get_sg_table,
     .gem_prime_vmap         = nv_drm_gem_prime_vmap,
     .gem_prime_vunmap       = nv_drm_gem_prime_vunmap,
-
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 0, 0)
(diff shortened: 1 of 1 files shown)


kernel-open/nvidia-drm/nvidia-drm-drv.c
```diff
@@ -826,7 +826,9 @@ static struct drm_driver nv_drm_driver = {
     .gem_prime_get_sg_table = nv_drm_gem_prime_get_sg_table,
     .gem_prime_vmap         = nv_drm_gem_prime_vmap,
     .gem_prime_vunmap       = nv_drm_gem_prime_vunmap,
-
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 0, 0)
```
mtijanic, line 829: I know this still says WIP, but take a look at this related conftest instead of kernel version check:
https://github.com/NVIDIA/open-gpu-kernel-modules/blob/main/kernel-open/conftest.sh#L4342-L4364

mtijanic (commented):
Thanks for the PR! I won't look too closely while it still has the WIP tag, but I'm curious, what are some particular usecases you are targeting with this change?

And related, what would be good ways to test and validate the change?

YusufKhan-gamedev:
The usecase is for whenever a user wants to convert a GEM handle to a dmabuf handle and then mmap it. I couldnt get a example working on my RADEON laptop hence why this is marked WIP.

PR #368, opened 2022-09-12
Minor changes to `README`

obj-obj:


mtijanic:
Hi and thank you for the interest. Per https://github.com/NVIDIA/open-gpu-kernel-modules/blob/main/CONTRIBUTING.md we are currently not accepting cosmetic PRs:
> While we appreciate your enthusiasm, we have decided not to accept non-functional changes such as non-code typo fixes, comment and language adjustments, whitespace changes, and similar.
> 
> Changes going into this codebase incur significant overhead. As such, we want to focus our resources on executable code improvements for now.

I'll mark this PR as `cosmetic-postponed` for now.

FYI, README.md is generated, so any changes actually would have to be made on the internal source files instead, and then integrated across our internal driver branching scheme, each with its own CI unit tests run. And then merged with this commit when doing a code drop to preserve authorship.

PR #373, opened 2022-09-15
Beginner's Instruction

cabelo:


cabelo:
@mmaneetsingh is ok?

carlosedp:
It would be good to add the requirements to compile like `build-essential` on Ubuntu, and etc.

FunPineapple:
Beginners should learn how Git works before attempting to compile from source. If repos are responsible for distributing common and readily-available knowledge, NVIDIA might as well put guides on installing every distro in the Readme as well. The instructions already provided are not common and are unique to the kernel modules.

cabelo:
> Beginners should learn how Git works before attempting to compile from source. If repos are responsible for distributing common and readily-available knowledge, NVIDIA might as well put guides on installing every distro in the Readme as well. The instructions already provided are not common and are unique to the kernel modules.

you are right.

PR #493, opened 2023-04-20
Refactor always-false conditional

meme8383:
Fix always false expression checking return value of write to adapter->name


ryao:
Why did Nvidia's coverity scans not catch this? I would have expected them to spot this. Did someone just not bother to look at the report?

aritger:
The 545.23.06 release contains a fix for this bug, though it was fixed in a slightly different way than the proposal here.  Thanks again for pointing this out.

Issue #433, opened 2022-12-30
Open-source kernel modules are not working


Monsterovich:
### NVIDIA Open GPU Kernel Modules Version

525.60.11

### Does this happen with the proprietary driver (of the same version) as well?

No

### Operating System and Version

Linux Mint 21.1

### Kernel Release

Linux pc 5.19.15-xanmod1 #0~20221012.git85103f7 SMP Wed Oct 12 15:04:07 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### Hardware: GPU

GPU 0: NVIDIA GeForce RTX 3060 Ti (UUID: GPU-039a34f7-087c-dc8f-a607-e0d937443d0f)

### Describe the bug

The kernel modules are loaded, but the GPU is not detected.

`dmesg` is filled with the same errors:

```
[  217.545140] NVRM nvAssertFailedNoLog: Assertion failed: status == NV_OK @ kernel_gsp_ga102.c:235
[  217.545142] NVRM kgspInitRm_IMPL: cannot bootstrap riscv/gsp: 0xffff
[  217.545145] NVRM RmInitAdapter: Cannot initialize GSP firmware RM
[  217.545832] NVRM: GPU 0000:05:00.0: RmInitAdapter failed! (0x62:0xffff:1622)
[  217.546288] NVRM: GPU 0000:05:00.0: rm_init_adapter failed, device minor number 0
[  217.729940] NVRM s_executeBooterUcode_TU102: Booter failed with non-zero error code: 0xa
[  217.729943] NVRM kgspExecuteBooterUnloadIfNeeded_TU102: failed to execute Booter Unload: 0xffff
[  217.738211] NVRM s_executeFwsec_TU102: failed to execute FWSEC for FRTS: FRTS error code 0xbe
[  217.738213] NVRM nvAssertOkFailedNoLog: Assertion failed: Failure: Generic Error [NV_ERR_GENERIC] (0x0000FFFF) returned from kgspExecuteFwsecFrts_HAL(pGpu, pKernelGsp, pKernelGsp->pFwsecUcode, pKernelGsp->pWprMeta->frtsOffset) @ kernel_gsp_ga102.c:164
```

### To Reproduce

```
sudo apt install nvidia-driver-525-open
```

### Bug Incidence

Always

### nvidia-bug-report.log.gz

[...]

PR #495, opened 2023-04-30
Show in README how to make this work on default Ubuntu install

majal:
This will address the following:
https://github.com/NVIDIA/open-gpu-kernel-modules/issues/433 https://answers.launchpad.net/ubuntu/+source/ubuntu-drivers-common/+question/705811


Binary-Eater:
Thank you. We no longer need the `NVreg_OpenRmEnableUnsupportedGpus` regkey.

https://github.com/NVIDIA/open-gpu-kernel-modules/blob/df1c9a3de2be4b130f4fc7680f7198193d8ac6ad/kernel-open/nvidia/nv-reg.h#L831-L838

PR #515, opened 2023-06-07
Hide compatible GPUs under a dropdown

edisionnano:
Not much to say here, makes the main page shorter.


niv:
Hello! Thank you for your PR. While this is a good UX change, as explained in [CONTRIBUTING.md](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/main/CONTRIBUTING.md), we do not currently accept cosmetic changes.

PR #549, opened 2023-08-22
fix: log message typo

0o001:


Binary-Eater:
Thank you for your contribution!

https://github.com/NVIDIA/open-gpu-kernel-modules/blob/df1c9a3de2be4b130f4fc7680f7198193d8ac6ad/src/common/displayport/src/dp_connectorimpl.cpp#L923

PR #560, opened 2023-09-17
nv-msi: Fix free_irq index on nv_request_msix_irq

guixiongwei:
In the nv_request_msix_irq, if the request_threaded_irq failed. The free_irq will be called to free the allocated irqs. The allocated irqs index should be j instead of i.


ttabi:
Fixed in 545.29.02

PR #567, opened 2023-10-31
Align DKMS package name with RPM spec

cartermckinnon:
The RPM spec for the kernel module DKMS package uses `nvidia-open` as the package name, but the `dkms.conf` in this repository uses `nvidia`.

If you:
```
yum install https://developer.download.nvidia.com/compute/cuda/repos/rhel7/x86_64/kmod-nvidia-open-dkms-545.23.06-1.el7.x86_64.rpm
```
Then you end up with a package in your DKMS tree called `nvidia-open`. If you `mktarball` and `ldtarball` that package, it becomes `nvidia`.


aritger:
Thanks for pointing this out.  I'm unsure if the `nvidia-open` in kmod-nvidia-open-dkms-545.23.06-1.el7.x86_64.rpm is correct.  I'll check with the maintainer of that package.

cartermckinnon:
FWIW, It's convenient in my case to use `nvidia-open`, to distinguish it from the closed-source kmod in my DKMS tree.

cartermckinnon:
@aritger do you have any update on this? I'd like to get rid of the `sed` I have to work around this 😄

aritger:
Sorry, I didn't get much traction on this before.  I've filed NVIDIA internal bug 4360392 for this, though.

cartermckinnon:
Sounds good, thanks for looking into it!

Issacwww:
Hey @aritger, is there any update for this?

aritger:
Sorry, we haven't made any progress on this, yet.

cartermckinnon:
I'm just going to close this to clear out my dashboard; but I believe this is still an issue.

PR #577, opened 2023-11-22
Makefile: Use $(MODLIB)/build/source symlink for kernel source

ErichRitz:
After kernel commit d8131c2965d5ee59bfa4d548641e52a13cbe17c9 ("kbuild: remove $(MODLIB)/source symlink") https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d8131c2965d5ee59bfa4d548641e52a13cbe17c9 $(MODLIB)/source no longer exists.  Add support for $(MODLIB)/build/source.

### DIFF
--- a/kernel-open/Makefile
+++ b/kernel-open/Makefile
@@ -28,7 +28,7 @@ else
   else
     KERNEL_UNAME ?= $(shell uname -r)
     KERNEL_MODLIB := /lib/modules/$(KERNEL_UNAME)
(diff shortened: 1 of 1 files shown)


ErichRitz:
See https://forums.developer.nvidia.com/t/modlib-source-symlink-no-longer-created-for-kernel-6-6-breaks-dkms-build/273723

aritger:
Thanks for the report of the problem and the proposed fix.  I've filed NVIDIA-internal bug 4390208 for this, and we'll try to incorporate your proposed fix in a future release.  Thanks.

Issue #574, opened 2023-11-11
crypto_shash_ctx_aligned has been removed from the linux kernel in the linux-next version


babariviere:
### NVIDIA Open GPU Kernel Modules Version

545.29.02 (sha256-RncPlaSjhvBFUCOzWdXSE3PAfRPCIrWAXyJMdLPKuIU=)

### Operating System and Version

NixOS 23.05

### Kernel Release

Linux 6.6.0-next-20231110 #1-NixOS SMP PREEMPT_DYNAMIC Fri Nov 10 01:06:34 UTC 2023 x86_64 GNU/Linux

### Build Command

```
make -C kernel-open modules
```

### Terminal output/Build Log

```
make -C kernel-open modules
make[1]: Entering directory '/build/source/kernel-open'
make[2]: Entering directory '/nix/store/8vmsmjxz2c547cgmmgnrzava7gglmqhs-linux-6.6.0-next-20231110-dev/lib/modules/6.6.0-next-20231110/source'
make[3]: Entering directory '/nix/store/8vmsmjxz2c547cgmmgnrzava7gglmqhs-linux-6.6.0-next-20231110-dev/lib/modules/6.6.0-next-20231110/build'
  SYMLINK /build/source/kernel-open/nvidia/nv-kernel.o
  SYMLINK /build/source/kernel-open/nvidia-modeset/nv-modeset-kernel.o
 CONFTEST: hash__remap_4k_pfn
 CONFTEST: set_pages_uc
 CONFTEST: list_is_first
 CONFTEST: set_memory_uc
 CONFTEST: set_memory_array_uc
 CONFTEST: set_pages_array_uc
 CONFTEST: ioremap_cache
 CONFTEST: ioremap_wc
 CONFTEST: ioremap_driver_hardened
 CONFTEST: ioremap_driver_hardened_wc
 CONFTEST: ioremap_cache_shared
 CONFTEST: pci_get_domain_bus_and_slot
 CONFTEST: pde_data
 CONFTEST: get_num_physpages
 CONFTEST: xen_ioemu_inject_msi
 CONFTEST: phys_to_dma
 CONFTEST: get_dma_ops
 CONFTEST: dma_attr_macros
 CONFTEST: dma_map_page_attrs
 CONFTEST: write_cr4
 CONFTEST: of_find_node_by_phandle
 CONFTEST: of_node_to_nid
 CONFTEST: pnv_pci_get_npu_dev
[...]

PR #589, opened 2024-01-11
Changed calls of crypto_tfm_ctx_aligned due to it's exclusion in kernels 6.7.0 or above

Emimendoza:
Since crypto_tfm_ctx_aligned got removed in kernel 6.7.0 (https://github.com/torvalds/linux/commit/acd7799574e57f1e494a5b85741eee78d1e93aca), we need to add it back in (or change the two uses of it) to be able to compile with newer kernels. 

Fixes issue: #574

Edit: Instead of re implementing (like my first commit) a better solution is to change the calls of crypt_ftm_ctx_aligned into doing what crypto_ftm_ctx_aligned used to do, that way this solution doesn't relay on conditional compilation and is backwards compatible.

### DIFF
--- a/kernel-open/nvidia/internal_crypt_lib.h
+++ b/kernel-open/nvidia/internal_crypt_lib.h
@@ -68,7 +68,6 @@
 #endif
 
 #ifdef USE_LKCA
-#include <linux/version.h>
 #include <linux/crypto.h>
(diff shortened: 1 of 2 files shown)


Septias:
What is blocking this?

svenstaro:
This is already in the new driver release.

PR #658, opened 2024-06-06
Patches for testing r555 stutter issues

mtijanic:
# This is not a pull request.

Just publishing my side branch here to have a centralized place to test and discuss these patches.

We've been having reports of stutter issues in 555 releases related to GSP enablement. On the proprietary driver, `NVreg_EnableGpuFirmware=0` makes them go away; on the open driver that's not an option.

So far, we've identified two possible causes here. One is fixed by commit 674c009 below. The other we can't fix/workaround in the kernel modules and requires usermode changes, but commit 8c1c49b should tell us if that path is actually being hit or not.

I've also augmented the logs captured by nvidia-bug-report.sh with some of the info that we found severely lacking in the bug reports so far.

My hope is that folks that have experienced these stutter issues can take these patches, try to reproduce the issue and report back with their findings (and their nvidia-bug-report logs). Many thanks in advance to anyone willing to go the extra mile(s) for us here!

We've unfortunately missed beta2 / 555.52 with this stuff (security fixes can't wait), but here it is early so we don't have to wait on the next release.

### DIFF
--- a/kernel-open/nvidia/nv.c
+++ b/kernel-open/nvidia/nv.c
@@ -4042,6 +4042,16 @@ int NV_API_CALL nv_get_event(
     nvidia_event_t *nvet;
     unsigned long eflags;
 
+    //
+    // Note that the head read/write is not atomic when done outside of the
+    // spinlock, so this might not be a valid pointer at all. But if we read
+    // NULL here that means that the value indeed was NULL and we can bail
+    // early since there's no events. Otherwise, we have to do a proper read
+    // under a spinlock.
+    //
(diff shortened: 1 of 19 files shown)


bkdwt:
@mtijanic Thanks for this patch. There's a possibility to your update your patch for support 550.100 drivers? 555.100 drivers fix a bug that caused OpenGL triple buffering to behave like double buffering. 

Thanks in advance! :)

ptr1337:
> Does that patch is needed for 560

They are already merged in 560.

PR #695, opened 2024-08-26
[backport] Prevent undefined behavior

Crispy-fried-chicken:
Here is a vulnerability CVE-2023-32690 in [libspdm](https://github.com/DMTF/libspdm) and you fix it in the main branch and 550 branch. Maybe it need to be backported to the 545.23 branch?


steven-bellock:
SPDM is only used for confidential compute and 545 is not used / endorsed for confidential compute, so the code path should never be exercised.

PR #715, opened 2024-10-02
Fix broken ELD files with mangled `monitor_name` when using HDMI

C0rn3j:
The Nvidia driver, both open and proprietary, currently pads the `monitor_name` of HDMI devices with newlines and extra spaces in the generated ELDs.   

This ends up breaking both PipeWire and Pulseaudio at minimum, see the [downstream report](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/4332) which is worked around by [this PipeWire PR](https://gitlab.freedesktop.org/pipewire/pipewire/-/merge_requests/2139).

The issue is easily verified by checking either the ELD files directly, or `device.product.name` from Pulse tools (provided they do not contain the workaround patch above), which rely on the ELD.

Based on what I could find by scourging public pastebins, and asking people around for outputs, only Nvidia has this issue, and only for HDMI device product names.

Pre-patch `cat /proc/asound/NVidia/eld\#0.*` which have a newline and spaces in them:
image  
*Line 7, which is only there due to the extra `\n`, has the extra spaces*

Pre-patch:
```bash
[130] % pactl --format=json list cards | python -m json.tool | grep '\\n'
                    "device.product.name": "27QHD240\n    "
                    "device.product.name": "ASUS MG279\n  "
```

Post-patch:
```bash
[0] % pactl --format=json list cards | python -m json.tool | grep 'ASUS'
                    "device.product.name": "ASUS MG279"
```

Out of curiosity, how long has this been in the driver?  

I can only tell that it's been in the open source init that goes back 2 years.

---

*I am not a C developer, please have mercy if there is some glaring issue.*

*I am assuming my commits will be squashed before merge, they're a mess.*

### DIFF
--- a/src/nvidia-modeset/src/nvkms-hdmi.c
+++ b/src/nvidia-modeset/src/nvkms-hdmi.c
@@ -1032,6 +1032,30 @@ static NvBool FillELDBuffer(const NVDpyEvoRec *pDpyEvo,
      * string in the EDID is terminated with 0x0A and padded with 0x20.
      * We do not keep the special characters.
      */
+    if (status == NVT_STATUS_SUCCESS) {
+        /*
+         * NvTiming_GetProductName() returns a nul-terminated string, but the
+         * string in the EDID is terminated with 0x0A and padded with 0x20.
+         * Put back these special characters.
+         */
+//        NvBool pastTerminator = FALSE;
+        NvU32 i;
+
+//        for (i = 0; i < NVT_EDID_LDD_PAYLOAD_SIZE; i++) {
+//            if (pastTerminator) {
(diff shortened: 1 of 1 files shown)


aaronp24:
Thanks for reporting this. While it's not obvious from the Intel HDA spec what the format of this string is _supposed_ to be, it doesn't hurt to trim it to match what other drivers do.

I ended up merging a slightly different version of the change that will end up in a future release. In the meantime, I pushed a version of it here:
https://github.com/NVIDIA/open-gpu-kernel-modules/compare/main...aaronp24:open-gpu-kernel-modules:eld-trim-monitor-name

Issue #732, opened 2024-11-11
The cudaHostRegister call fails on page-backed memory mapped to userspace with remap_pfn_range since version 550.54.14.


lizhengui007:
### NVIDIA Open GPU Kernel Modules Version

550.54.14

### Please confirm this issue does not happen with the proprietary driver (of the same version). This issue tracker is only for bugs specific to the open kernel driver.

- [ ] I confirm that this does not happen with the proprietary driver package.

### Operating System and Version

Rocky Linux release 8.10 (Green Obsidian）

### Kernel Release

Linux virtaitech-hz-ws2 4.18.0-553.16.1.el8_10.x86_64 #1 SMP Thu Aug 8 17:47:08 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

### Please confirm you are running a stable release kernel (e.g. not a -rc). We do not accept bug reports for unreleased kernels.

- [ ] I am running on a stable kernel release.

### Hardware: GPU

GPU 0: NVIDIA GeForce GTX 1050 Ti (UUID: GPU-fd200267-5107-2f45-63f6-24848e1ac6c6)

### Describe the bug

I found a problem that  the cudaHostRegister call fails on page-backed memory mapped to userspace with remap_pfn_range since version 550.54.14.
Previous versions such as 550.40.07 or earlier versions can be executed successfully.  The err return by cudaHostRegister is “invalid argument”. I don't know why this change happened from version  550.54.14. Doesn’t the new version code support registration for pages generated by remap_pfn_range?

### To Reproduce

I provide a sample code  to reproduce it.  Please decompress this compressed package cudaHostRegister_demo.tar.gz

The sample code is divided into kernel module (kernel directory) and user programs (user directory).
1、Compile the kernel module.
   Enter the kernel directory: make -C /lib/modules/$(uname -r)/build M=`pwd` modules, and then insmod my_misc_device.ko.
   The kernel part provides a mmap interface for mapping a kernel page to user space using remap_pfn_range.

2、Compile the user programs.
   Enter the user directory:  make and generate cudaHostRegister_demo，and then execute it.  The execution of  cudaHostRegister_demo return error.

The result of my test demo is below:
[root@virtaitech-hz-ws2 user]# ./cudaHostRegister_demo
normal map memory mapped at address 0x7f023282e000
normal_map cudaHostRegister success 
remap_pfn_range memory mapped at address 0x7f0232649000
[...]

PR #742, opened 2024-11-18
Fix cudaHostRegister failed on page-backed memory mapped to userspace with remap_pfn_range

lizhengui007:
Fixes issue: https://github.com/NVIDIA/open-gpu-kernel-modules/issues/732


lizhengui007:
Could anyone give me a reply?

sirenuf:
so much for open source

aritger:
I apologize you never received a response.

Unfortunately, this path needs some work:

- nv_follow_pfn() uses either follow_pfn() or follow_pte(), based on conftest.
- In modern, kernels, both of those have been removed.  So, nv_follow_pfn() always returns -1.

We plan to rework this path to use an alternate mechanism.

PR #783, opened 2025-02-13
removing .date from nvidia-drm-drv.c

nbanb:
error: ‘struct drm_driver’ has no member named ‘date’

DATE field had been totally removed from mainline Linux kernel (linux-6.14-rc). 
See commit https://github.com/torvalds/linux/commit/cb2e1c2136f71618142557ceca3a8802e87a44cd 
This change had to be done on proprietary drivers too (e.g nvidia-550.144.03)


Binary-Eater:
Thanks for contribution! This has been resolved since 570.124.04

https://github.com/NVIDIA/open-gpu-kernel-modules/commit/129479b1b7345f0055016ab3aa42ae78a4ca2048

PR #792, opened 2025-02-21
prefer dkms ${kernelver} over uname -r resolves #791

### LINKED ISSUE #791: dkms incorrectly builds updated kernels with the current kernel's source
### NVIDIA Open GPU Kernel Modules Version

nvidia-open/570.86.15

### Operating System and Version

Fedora 41

### Kernel Release

Linux d0 6.12.15-200.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Tue Feb 18 15:24:05 UTC 2025 x86_64 GNU/Linux

### Please confirm you are running a stable release kernel (e.g. not a -rc). We do not accept bug reports for unreleased kernels.

- [x] I am running on a stable kernel release.

### Build Command

`dnf upgrade` (which invokes `dkms build` under the running kernel)

### Terminal output/Build Log

Notice it enter's the 'wrong' `/usr/src/kernels` directory,  ie:

```
DKMS (dkms-3.1.5) make.log for nvidia-open/570.86.15 for kernel 6.12.15-200.fc41.x86_64 (x86_64)
Fri Feb 21 09:50:50 AM EST 2025
Cleaning build area
# command: 'make' clean
make -C src/nvidia clean
make[1]: Entering directory '/var/lib/dkms/nvidia-open/570.86.15/build/src/nvidia'
rm -f -rf _out/Linux_x86_64
make[1]: Leaving directory '/var/lib/dkms/nvidia-open/570.86.15/build/src/nvidia'
make -C src/nvidia-modeset clean
make[1]: Entering directory '/var/lib/dkms/nvidia-open/570.86.15/build/src/nvidia-modeset'
rm -f -rf _out/Linux_x86_64
make[1]: Leaving directory '/var/lib/dkms/nvidia-open/570.86.15/build/src/nvidia-modeset'
make -C kernel-open clean
make[1]: Entering directory '/var/lib/dkms/nvidia-open/570.86.15/build/kernel-open'
rm -f -r conftest
make[2]: Entering directory '/usr/src/kernels/6.12.10-200.fc41.x86_64'
make[2]: Leaving directory '/usr/src/kernels/6.12.10-200.fc41.x86_64'
make[1]: Leaving directory '/var/lib/dkms/nvidia-open/570.86.15/build/kernel-open'
...
```

### More Info

I landed here because of `nvidia-open/570.86.15` issues on Fedora 41.  When `dnf upgrade` is run, and it installs a new kernel, then `dkms` will automatically build `nvidia-open` under the running kernel (without explicitly setting a value for `KERNEL_UNAME`).   The currently included [Makefile](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/main/kernel-open/Makefile) is being used.

That is evident in the `/var/lib/dkms/nvidia-open/*/*/*/log/make.log` file(s).  

Subsequent reboots fail to load the module, eg:

```
kernel: nvidia: version magic '6.12.10-200.fc41.x86_64 SMP preempt mod_unload ' should be '6.12.11-200.fc41.x86_64 SMP preempt mod_unload '
```

Here's a short work-around (for Fedora) that leverages `KERNEL_UNAME`  and properly build `nvidia-open` for the 'latest' kernel.

After a `dnf upgrade`, a variation of following will work as `root`.

```sh
# do this as root
sudo su -

# fyi, version sort installed kernels
rpm -qa kernel | sed -e 's/^kernel-//g' | sort -uV

# fyi, version sort installed dkms module/module-version and kernel/arch
dkms status | sort -uV

export CURRENT_KERNEL="$(uname -r)"; echo "CURRENT_KERNEL=${CURRENT_KERNEL}"
export LATEST_KERNEL="$(rpm -qa kernel | sed -e 's/^kernel-//g' | sort -uV | tail -1)"; echo LATEST_KERNEL=${LATEST_KERNEL} # this matches uname -r

# example rebuild; pick one
export KERNEL_UNAME="6.12.11-200.fc41.x86_64"
export KERNEL_UNAME=${CURRENT_KERNEL}
export KERNEL_UNAME=${LATEST_KERNEL}
echo KERNEL_UNAME=${KERNEL_UNAME}

# set proper values for dkms build, install, etc
export DKMS_ARCH="$(dkms status | grep ${KERNEL_UNAME}, | awk -F, '{print $3}' | awk '{print $1}' | awk -F: '{print $1}')"
export DKMS_KERNEL="$(dkms status | grep ${KERNEL_UNAME}, | awk -F, '{print $2}' | awk '{print $1}')" # should be the same as KERNEL_UNAME
export DKMS_MODULE_VERSION="$(dkms status | grep ${KERNEL_UNAME}, | awk -F, '{print $1}' | awk '{print $1}')"

# manually verify values
echo DKMS_ARCH=${DKMS_ARCH}
echo DKMS_KERNEL=${DKMS_KERNEL}
echo DKMS_MODULE_VERSION=${DKMS_MODULE_VERSION}

# NOTICE! Using the LATEST_KERNEL value is easiest/safest.
# NOTICE! If you're booted with the latest kernel and the modules ARE NOT loaded, then properly rebuilding may immediately load the correct signed module and will likely reset a graphical session.
# IMPORTANT! If you're booted from the kernel you want to 'fix' then do this in a tmux, screen, or from the linux console.
KERNEL_UNAME=${DKMS_KERNEL} dkms uninstall ${DKMS_MODULE_VERSION} -k ${DKMS_KERNEL}/${DKMS_ARCH}
KERNEL_UNAME=${DKMS_KERNEL} dkms build ${DKMS_MODULE_VERSION} -k ${DKMS_KERNEL}/${DKMS_ARCH} --force
KERNEL_UNAME=${DKMS_KERNEL} dkms install ${DKMS_MODULE_VERSION} -k ${DKMS_KERNEL}/${DKMS_ARCH}
KERNEL_UNAME=${DKMS_KERNEL} dkms status ${DKMS_MODULE_VERSION} -k ${DKMS_KERNEL}/${DKMS_ARCH}

# verify the build make.log enters the correct /usr/src/kernels directory ...
less /var/lib/dkms/nvidia-open/*/${DKMS_KERNEL}/${DKMS_ARCH}/log/make.log

systemctl reboot
```

Using `uname -r` in the `Makefile` seems a  bit too convenient for a dynamic kernel module.

In the meantime, I posted the workaround here hoping it would help.  Thanks

josephtingiris:


aritger:
Thanks for the proposed change, @josephtingiris .

But, isn't mapping from dkms' `${kernelver}` variable to `KERNEL_UNAME` a more suitable job for dkms.conf?  This Makefile is generally not dkms-aware and already provides KERNEL_UNAME as the API to specify the target kernel.

In fact, it looks like dkms.conf already does `KERNEL_UNAME=${kernelver}`:
https://github.com/NVIDIA/open-gpu-kernel-modules/blob/main/kernel-open/dkms.conf#L9

Is that not working for some reason in your configuration?

josephtingiris:
> Thanks for the proposed change, @josephtingiris .
> 
> But, isn't mapping from dkms' `${kernelver}` variable to `KERNEL_UNAME` a more suitable job for dkms.conf? This Makefile is generally not dkms-aware and already provides KERNEL_UNAME as the API to specify the target kernel.
> 
> In fact, it looks like dkms.conf already does `KERNEL_UNAME=${kernelver}`: https://github.com/NVIDIA/open-gpu-kernel-modules/blob/main/kernel-open/dkms.conf#L9
> 
> Is that not working for some reason in your configuration?

No, that dkms.conf doesn't work and I noted why, here: https://github.com/NVIDIA/yum-packaging-nvidia-driver/issues/10#issuecomment-2675447010

IMO, the Makefile change is a way to keep the logic & functionality contained & maintained in a single location.  There are already multiple dkms.conf files in play (also noted in the link above).  Keep it simple.

The logic I added will preserve backward compatibility. $KERNEL_UNAME from the environment will still be honored and take precedence unless $kernelver is also present.  If both are present then it tests for equality and will error if they don't match (which is how I found the issue to begin with).  That seems more apropos.  I thought it was better/safer behavior for make to error than to compile a module that fails to load on the next reboot (and renders the system significantly less useful without its gui).  In my testing, the modules compile just fine but the version magic doesn't match, so it won't load.

aritger:
I'm sorry, but the Makefile here isn't dkms-aware, and I don't believe it should be dkms-aware.  If there are problems with dkms.conf, let's please fix that.

josephtingiris:
No need.  Happy to help if I can.  I previously outlined fixing dkms.conf in the packaging repo, too.  Those repos are quite dated.

The kernel-open/dkms.conf that's in this repo looks like a template for something else.  It's also 3 years old.  By itself, it doesn't function.  Though it could easily be made to.  I see other bits of dkms in here, too, though not in the Makefile.   Are you implying the dkms.conf here get fixed?

Also, I searched and reviewed all of https://github.com/search?q=org%3ANVIDIA%20dkms.conf&type=code and the dkms.conf that's included in the rpm doesn't match any of them.  I also searched for matching content and can't find it.  I'll keep looking but I'm not sure how one would effect a change to that particular dkms.conf.  I guess it's in a private repo.

Thanks for your time.

1Naim:
> The kernel-open/dkms.conf that's in this repo looks like a template for something else. It's also 3 years old

Yes you're right. It's a template from the NVIDIA runfile. The runfile will update `dkms.conf` as needed as per [nvidia-installer/files.c](https://github.com/NVIDIA/nvidia-installer/blob/main/files.c) but that's obviously not an option for packagers and I assume this repo too. 

At least in Arch, they're making the modifications during the packaging state so `dkms.conf` is useable. See [1] and [2]

[1] [dkms.conf patching for the closed kernel modules](https://gitlab.archlinux.org/archlinux/packaging/packages/nvidia-utils/-/blob/main/PKGBUILD?ref_type=heads#L63-75)
[2] [dkms.conf patching for the open kernel moduels](https://gitlab.archlinux.org/archlinux/packaging/packages/nvidia-utils/-/blob/main/PKGBUILD?ref_type=heads#L63-75) -- Also dkms.conf is [moved](https://gitlab.archlinux.org/archlinux/packaging/packages/nvidia-utils/-/blob/main/PKGBUILD?ref_type=heads#L343) to the top directory because otherwise it won't build due to missing targets.

PR #843, opened 2025-04-30
Add detection of loongarch64 CPUs

vasthecat:
Code in `cpyopsys.h` would incorrectly assume that architecture is 32-bit if built on loongarch64 machine. This could lead to errors for users of these headers.


Binary-Eater:
Thank you for your contribution @vasthecat. This driver implementation is designed currently to be supported by `x86_64` and `aarch64` architectures. Similarly, we do not qualify building the drivers on `loongarch64` build systems. Any other CPU architecture support you see in the code base comes from support we had done previously, but no longer continue to support.

PR #940, opened 2025-10-01
Allow custom values for KERNEL_MODLIB

renhiyama:
Currently, as far as I know - there's no way to modify this variable to force make script to look for custom directory where linux kernel was built. Right now, nvidia drivers are hardcoded to look at `/lib/modules/<kernel_uname>`, and doesn't have option for tinkerers like me to point to custom kernel builds.

A valid reason from my side would be that I am currently building kernel without root permissions. and so I can't install it in `/lib` folder, but use a temporary folder to currently compile nvidia driver too.

### DIFF
--- a/kernel-open/Makefile
+++ b/kernel-open/Makefile
@@ -27,7 +27,7 @@ else
     KERNEL_SOURCES := $(SYSSRC)
   else
     KERNEL_UNAME ?= $(shell uname -r)
-    KERNEL_MODLIB := /lib/modules/$(KERNEL_UNAME)
+    KERNEL_MODLIB ?= /lib/modules/$(KERNEL_UNAME)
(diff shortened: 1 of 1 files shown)


aritger:
Thanks for the pull request.  But, does setting SYSSRC already achieve the same thing?
```
  ifdef SYSSRC
    KERNEL_SOURCES := $(SYSSRC)
  else
    KERNEL_UNAME ?= $(shell uname -r)
    KERNEL_MODLIB := /lib/modules/$(KERNEL_UNAME)
    KERNEL_SOURCES := $(shell ((test -d $(KERNEL_MODLIB)/source && echo $(KERNEL_MODLIB)/source) || (test -d $(KERNEL_MODLIB)/build/source && echo $(KERNEL_MODLIB)/build/source)) || echo $(KERNEL_MODLIB)/build)
  endif
```

renhiyama:
Oh, I just checked and yes - I can definitely use SYSSRC & SYSOUT variables!
At first I thought that maybe SYSSRC is used in some other code too, but just ran grep through the repo and it seems this is the only instance! 

 Extremely sorry! Yes, SYSSRC & SRCOUT is definitely better for my problem - I should have paid attention to those ones 😅

PR #948, opened 2025-10-13
mapping_reuse: correct range size when unmapping failed allocations

mateusz-bloch:
Fixes incorrect unmap range size in the error cleanup path of reusemappingdbMap().

Previously, range.size was used instead of the actual mapped fragment size (token.pList->size) when calling pUnmapCb.   This could result in over-unmapping and overlapping VA ranges, causing inconsistencies and VA allocation failures.


moinak-nv:
This fix has been integrated locally in our development builds and will make it out in a future release. Thank you for your contribution.

aritger:
Thanks for the fix, @mateusz-bloch .  Closing.

PR #984, opened 2025-12-11
nvidia: add RmForceExternalGpu registry key

roger-pmta:
Thunderbolt 4/5 eGPU enclosures fail automatic external GPU detection
because they are not in the approved TB3 bridge list and do not report
the required slot capabilities (`HotPlug+` and `Surprise+`).

Add `RmForceExternalGpu` registry key to bypass bridge detection and
unconditionally set `PDB_PROP_GPU_IS_EXTERNAL_GPU,` enabling proper eGPU
handling throughout the driver.

**Usage:**

Global (all GPUs):

      NVreg_RegistryDwords="RmForceExternalGpu=1"

Per-device (specific GPU only):

      NVreg_RegistryDwordsPerDevice="pci=DDDD:BB:DD.F;RmForceExternalGpu=1"

The per-device option allows targeting specific GPUs in multi-GPU systems.

### DIFF
--- a/src/nvidia/arch/nvalloc/unix/src/osinit.c
+++ b/src/nvidia/arch/nvalloc/unix/src/osinit.c
@@ -1310,10 +1310,29 @@ RmInitNvDevice(
     }
 
     // Configure eGPU setting
-    if (RmCheckForExternalGpu(pGpu, pCl))
     {
-        pGpu->setProperty(pGpu, PDB_PROP_GPU_IS_EXTERNAL_GPU, NV_TRUE);
-        nv->is_external_gpu = NV_TRUE;
+        NvU32 forceExternalGpu = 0;
+        NvBool isExternalGpu = NV_FALSE;
+
+        // Check if user has forced eGPU mode via RmForceExternalGpu registry key
+        if ((osReadRegistryDword(pGpu, NV_REG_STR_RM_FORCE_EXTERNAL_GPU, &forceExternalGpu) == NV_OK) &&
+            (forceExternalGpu != 0))
+        {
+            NV_PRINTF(LEVEL_NOTICE, "Forcing external GPU mode via RmForceExternalGpu registry key\n");
+            isExternalGpu = NV_TRUE;
+        }
+        else
+        {
+            // Otherwise, use automatic detection
+            isExternalGpu = RmCheckForExternalGpu(pGpu, pCl);
+        }
+
+        if (isExternalGpu)
+        {
+            NV_PRINTF(LEVEL_NOTICE, "GPU treated as external GPU (eGPU)\n");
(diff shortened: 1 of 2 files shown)


roger-pmta:
Workaround for #979

roger-pmta:
Full write-up with 590 and TB5 posted at https://egpu.io/forums/builds/2023-14-lenovo-thinkpad-x1-carbon-gen-11-13th10cu-rtx-5080-32gbps-tb4-sonnet-breakaway-box-850-t5-linux-rocky-10-1/, will wait for a reply/update from NVIDIA.

mtijanic:
Hi. Sorry, I'm not really familiar with eGPU details to know what kind of effects such a change can invoke, but I can make a couple of comments:

I'm actually surprised that this works. You've declared a top level module param, but are reading it as if it's an RM registry dword. The typical way to read top level params is to just declare them as extern vars and read them directly, like e.g. [here](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/590.44.01/kernel-open/nvidia/nv-platform.c#L1378).

The `osReadRegistryDwords()` function is meant for reading keys as defined in [nvrm_registry.h](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/590.44.01/src/nvidia/interface/nvrm_registry.h), and those are supposed to be passed to the driver as `option nvidia NVreg_RegistryDwords="foo=1;bar=0x3f"`.

However, the registry dword is IMO the better thing to use here, for two reasons:
1. It can also be set per-GPU, by using `NVreg_RegistryDwordsPerDevice` and passing the PCI ID for which GPU to apply it to. See [here](https://github.com/NVIDIA/open-gpu-kernel-modules/blob/590.44.01/kernel-open/nvidia/nv-reg.h#L224-L252) for docs)
2. The registry dwords interface is a lot more experimental/unstable and is more suitable for these WARs. I can see us merging something like that, because it implies a lot less support than the top level params.

To use this, revert your `nv-reg.h` changes, and add your key to `nvrm_registry.h`. See other keys defined there for examples.

roger-pmta:
Thanks for the feedback @mtijanic! Updated the patch with your suggestions and confirmed working 🙌

bdandy:
I'm curious why you decided to make this as registry dwords, not kernel module option so it would be visible in modinfo?

orkineric:
Confirmed working on RTX 5090 + Gigabyte Aorus RTX 5090 AI Box (TB5) / Fedora 43 / kernel 7.0.0-rc7 alongside PR #1109, which makes the ReBAR resize non-fatal (and adds a `pci_is_thunderbolt_attached()` fast-path) so the nvidia driver actually binds the eGPU in the first place — without that, your `RmForceExternalGpu` registry key has no GPU to apply to on TB5. Both patches deployed together on a dual-5090 workstation (internal PCIe 5.0 x16 + Aorus AI Box over TB5), CUDA workloads including ollama multi-GPU inference across both cards running cleanly with no Xid errors.

PR #993, opened 2025-12-30
uvm: Fix the cleanup procedure after an error occurs

mapicccy:
After `nvUvmInterfaceSessionCreate` fails, the resources are not properly cleaned up, resulting in the inability to load the driver a second time


Binary-Eater:
Thank you for the change @mapicccy. We have integrated the change internally, accrediting you for the contribution. The change should make its way to this repository in future releases.

Issue #450, opened 2023-01-30
Black screen (monitor doesn't wake) after waking up from suspend


lufog:
### NVIDIA Open GPU Kernel Modules Version

525.85.05-2

### Does this happen with the proprietary driver (of the same version) as well?

No

### Operating System and Version

Arch Linux

### Kernel Release

6.1.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 24 Jan 2023 21:07:04 +0000 x86_64 GNU/Linux

### Hardware: GPU

NVIDIA GeForce GTX 1660

### Describe the bug

After wake up from suspend, the monitor does not wake up (no signal), the system unit wakes up, the coolers make noise, the keyboard backlight turns on. The indicator changes its state when the num-lock key is pressed, but only for the first 5-10 seconds, then the keyboard stops responding to pressing. The only way to get out of this state is to do a hard reset.

The problem appears on both Wayland and Xorg, but only on an open driver (nvidia-open), on a proprietary driver (nvidia) the computer wakes up without problems.

### To Reproduce

(It is enough to do the first two steps, reboot, and send the computer to sleep. But I will describe all the actions I have done.)

* Install `nvidia-open` package;
* remove `kms` from the `HOOKS` array in `/etc/mkinitcpio.conf` and regenerate the initramfs.
* add the `nvidia_drm.modeset=1` kernel parameter;
* create `/etc/modprobe.d/nvidia-power-management.conf` with content:
```
options nvidia NVreg_PreserveVideoMemoryAllocations=1 NVreg_TemporaryFilePath=/var/tmp
```
* enable services `nvidia-suspend.service`, `nvidia-hibernate.service`, `nvidia-resume.service`

### Bug Incidence

Always

### nvidia-bug-report.log.gz

[nvidia-bug-report.log.gz](https://github.com/NVIDIA/open-gpu-kernel-modules/files/10540124/nvidia-bug-report.log.gz)

### More Info

The problem and symptoms are the same as described by another user [here](https://bbs.archlinux.org/viewtopic.php?pid=2047692). Perhaps some of the information he provided will be useful.

Fxzzi:
I also experience this exact problem. RTX 3070 FE. Only occurs on open drivers too.

BurntToast01:
On a ryzen 2700x Kernel 6.1
I had the issue happen this morning after having rtcwake run. The PC woke up, but the screen remained black after returning from a suspended mode.
[...]

PR #996, opened 2026-01-04
nvidia-drm: Fix display not waking after suspend/resume

destenson:
Fixes display wake failure after system suspend/resume by restoring the display mode configuration on resume. Previously, drm_mode_config_reset() was called on suspend to clear display state, but no corresponding restore operation was performed on resume, leaving displays in a black screen state even though the system had successfully resumed.

This creates an asymmetry where:
- Suspend path: calls drm_mode_config_reset() to clear display config
- Resume path: did nothing, assuming VT switch would restore modes

When running X11/Wayland with a modeset owner, the NVKMS layer skips console restore (assuming the DRM client will handle it), but the DRM layer wasn't restoring modes (assuming VT switch would handle it). The result was that neither layer restored the display.

The fix adds drm_mode_config_reset() on the resume path to reinitialize the display outputs, matching the suspend behavior and ensuring displays are properly restored.

Tested on Ubuntu 24.04 with RTX 3090/3080, kernel 6.8.0-90-generic.

Fixes: https://github.com/NVIDIA/open-gpu-kernel-modules/issues/450

### DIFF
--- a/kernel-open/nvidia-drm/nvidia-drm-drv.c
+++ b/kernel-open/nvidia-drm/nvidia-drm-drv.c
@@ -2172,6 +2172,11 @@ void nv_drm_suspend_resume(NvBool suspend)
 
     /*
      * NVKMS shuts down all heads on suspend. Update DRM state accordingly.
+     *
+     * BUGFIX: Previously, drm_mode_config_reset() was called on suspend
+     * but nothing restored the display modes on resume, causing black screen
+     * after suspend/resume. Now we properly restore the mode config on resume.
+     * See: https://github.com/NVIDIA/open-gpu-kernel-modules/issues/450
      */
     for (nv_dev = dev_list; nv_dev; nv_dev = nv_dev->next) {
         struct drm_device *dev = nv_dev->dev;
@@ -2191,6 +2196,8 @@ void nv_drm_suspend_resume(NvBool suspend)
             drm_fb_helper_set_suspend_unlocked(dev->fb_helper, 0);
 #endif
             drm_kms_helper_poll_enable(dev);
+            /* Restore mode config to re-enable display output after resume */
+            drm_mode_config_reset(dev);
         }
     }
 


Binary-Eater:
@destenson thank you for taking the time to look into this. Our expectation is that we would not need to call `drm_mode_config_reset` both in the suspend and resume paths.

https://elixir.bootlin.com/linux/v6.18.3/source/drivers/gpu/drm/drm_mode_config.c#L184

In upstream DRM drivers, `drm_mode_config_reset` is only called for initialization. However, for nvidia-drm and nvidia-modeset, we need to call this on suspend since we suddenly shutdown the heads. We would expect a modeset to be performed on resume, so an explicit call to `drm_mode_config_reset` should not be needed. I suspect that this change might be masking some other bug that is affecting your system from resuming gracefully. If you have ssh configured on the system, would you mind running `sudo nvidia-bug-report.sh` without this change after attempting a resume from suspend?

Binary-Eater:
>When running X11/Wayland with a modeset owner, the NVKMS layer skips console restore (assuming the DRM client will handle it), but the DRM layer wasn't restoring modes (assuming VT switch would handle it). The result was that neither layer restored the display.

I would expect this to be solved by calling `pm_vt_switch_required`. We have an internal change to do this already, but I see it has not made its way to a public release yet.

destenson:
> > When running X11/Wayland with a modeset owner, the NVKMS layer skips console restore (assuming the DRM client will handle it), but the DRM layer wasn't restoring modes (assuming VT switch would handle it). The result was that neither layer restored the display.
> 
> I would expect this to be solved by calling `pm_vt_switch_required`. We have an internal change to do this already, but I see it has not made its way to a public release yet.

I had Claude Code investigate the problem & make the fix. It seemed like a very weird kind of fix, but it does seem to work.

Though I'm glad to hear it's been fixed already upstream. Hopefully it will be public & available in distributions soon. 

I'll close this PR after I run `sudo nvidia-bug-report.sh` this weekend with the unfixed version as you requested.

Thanks.

Binary-Eater:
@destenson here is the patch one of my coworkers worked on that I suspect could help resolve the issue you are seeing.

https://github.com/Binary-Eater/open-gpu-kernel-modules/commit/a5bd4df7811536d9efe7b084a34b47d3f6691ba2

If you wouldn't mind separate from the bug report collection, would you mind testing this patch out as well?

PR #997, opened 2026-01-05
Fix bashism in conftest.sh

shay-kl:
The script uses == for string comparison, which is a bashism. This causes failure on systems where /bin/sh is a strictly POSIX shell such as Dash. This fixes the syntax to use the standard = operator.

### DIFF
--- a/kernel-open/conftest.sh
+++ b/kernel-open/conftest.sh
@@ -5281,7 +5281,7 @@ case "$5" in
 
         if [ -n "$VGX_BUILD" ]; then
             if [ -f /proc/xen/capabilities ]; then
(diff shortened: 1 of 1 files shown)


aritger:
Thank you for the bug fix.  Did you actually encounter this bug, or was it found by inspection?  I think this would only have been triggered on special `VGX_BUILD` builds.

In any case, I've incorporated this change into NVIDIA's code base, and it will be included in a future driver release (for weird internal source control management reasons, I can't simply merge this pull request).

Thank you.

shay-kl:
I was switching my default shell to dash and inspected all my installed packages using `checkbashisms` for any potential issues. :)

Issue #971, opened 2025-11-22
nvidia-smi hangs indefinitely after ~66 days 12 hours uptime with driver 570.133.20 OpenRM on B200 and kernel 6.6.0


nickasam:
### NVIDIA Open GPU Kernel Modules Version

[root@A11-R42-I61-42-5504045 ~]# cat /proc/driver/nvidia/params ResmanDebugLevel: 4294967295 RmLogonRC: 1 ModifyDeviceFiles: 1 DeviceFileUID: 0 DeviceFileGID: 0 DeviceFileMode: 438 InitializeSystemMemoryAllocations: 1 UsePageAttributeTable: 4294967295 EnableMSI: 1 EnablePCIeGen3: 0 MemoryPoolSize: 0 KMallocHeapMaxSize: 0 VMallocHeapMaxSize: 0 IgnoreMMIOCheck: 0 EnableStreamMemOPs: 0 EnableUserNUMAManagement: 1 NvLinkDisable: 0 RmProfilingAdminOnly: 1 PreserveVideoMemoryAllocations: 0 EnableS0ixPowerManagement: 0 S0ixPowerManagementVideoMemoryThreshold: 256 DynamicPowerManagement: 3 DynamicPowerManagementVideoMemoryThreshold: 200 RegisterPCIDriver: 1 EnablePCIERelaxedOrderingMode: 0 EnableResizableBar: 0 EnableGpuFirmware: 18 EnableGpuFirmwareLogs: 2 RmNvlinkBandwidthLinkCount: 0 EnableDbgBreakpoint: 0 OpenRmEnableUnsupportedGpus: 1 DmaRemapPeerMmio: 1 ImexChannelCount: 2048 CreateImexChannel0: 0 GrdmaPciTopoCheckOverride: 0 RegistryDwords: "" RegistryDwordsPerDevice: "" RmMsg: "" GpuBlacklist: "" TemporaryFilePath: "" ExcludedGpus: ""

### Please confirm this issue does not happen with the proprietary driver (of the same version). This issue tracker is only for bugs specific to the open kernel driver.

- [ ] I confirm that this does not happen with the proprietary driver package.

### Operating System and Version

[root@A11-R42-I61-42-5504045 ~]# cat /etc/openeuler-release openeuler release 2.0 (LTS-SP2) [root@A11-R42-I61-42-5504045 ~]#

### Kernel Release

[...]

PR #1014, opened 2026-01-25
fix: Fix jiffies wrap in os_get_monotonic_time_ns() causing hangs after 66 days

ma-ts:
Fix for `nvidia-smi` hanging after approximately 66 days of uptime.

  The function `os_get_monotonic_time_ns()` in [`kernel-open/nvidia/os-interface.c`](kernel-open/nvidia/os-interface.c) uses `jiffies_to_timespec64(jiffies, &ts)` to obtain
  monotonic time. The `jiffies` counter is an unsigned 32-bit value that wraps at 2^32 ticks. At `HZ=750` this occurs after `2^32 / 750 / 86400 = 66.3 days`. When the wrap
  occurs, `jiffies_to_timespec64()` returns a near-zero value, causing time to appear to jump backwards.

  This breaks timeout comparisons throughout the driver. Code in [`thread_state.c`](src/nvidia/src/kernel/core/thread_state.c), [`locks.c`](src/nvidia/src/kernel/core/locks.c),
  and [`gpu_timeout.c`](src/nvidia/src/kernel/gpu/gpu_timeout.c) stores a start time and later checks if `currentTime >= startTime + timeout`. After the wrap, `currentTime` is
  suddenly much smaller than `startTime`, so these comparisons behave incorrectly and all operations appear to have timed out immediately.

  The fix replaces the jiffies-based implementation with `ktime_get_raw_ts64()`, which reads from hardware timers and provides a monotonic 64-bit nanosecond timestamp that won't
   wrap for centuries. This matches the implementation already used by `os_get_monotonic_time_ns_hr()` in the same file.

### DIFF
--- a/kernel-open/nvidia/os-interface.c
+++ b/kernel-open/nvidia/os-interface.c
@@ -701,15 +701,9 @@ NvU64 NV_API_CALL os_get_monotonic_time_ns_hr(void)
 
 NvU64 NV_API_CALL os_get_monotonic_time_ns(void)
 {
-#if defined(NV_JIFFIES_TO_TIMESPEC_PRESENT)
-    struct timespec ts;
-    jiffies_to_timespec(jiffies, &ts);
-    return (NvU64) timespec_to_ns(&ts);
-#else
     struct timespec64 ts;
-    jiffies_to_timespec64(jiffies, &ts);
+    ktime_get_raw_ts64(&ts);
     return (NvU64) timespec64_to_ns(&ts);
-#endif
 }
 
 NvU64 NV_API_CALL os_get_monotonic_tick_resolution_ns(void)


pjaroszynski:
jiffies should be 64-bit, right?

mtijanic:
> jiffies should be 64-bit, right?

Yeah, on 64bit systems (the only ones supported by this codebase), `jiffies` will be a regular 64bit variable, so this won't overflow. Also, as mentioned in https://github.com/NVIDIA/open-gpu-kernel-modules/issues/971#issuecomment-3798676875 the nvidia-smi hang bug is not part of this codebase at all. (Also, who uses `CONFIG_HZ=750`?)

That said, I think moving from `jiffies` to `ktime` here might be worthwhile anyway. However, AFAICT it was added in 4.18, and we still support 4.4 and later, so it would need some extra conftests to decide, and given that I don't think it worth it to have diverging behavior based on the kernel. Maybe wait until we drop those kernels and then move unconditionally....

tycho:
Why not just use `ktime_get_raw_ns` to eliminate the need for the `timespec64_to_ns` conversion?

duanyayong:
After the wrap, `currentTime` will suddenly smaller than `startTime`, then `currentTime` will smaller than `startTime + timeout` too. So, why is there a timeout occur in all operations?

PR #1018, opened 2026-02-03
nvidia-peermem: Prefer ofa_kernel-dkms path for DOCA-OFED headers

blochl:
New DKMS-based DOCA-OFED installations store headers in `/usr/src/ofa_kernel-dkms/` and create symlinks from `/usr/src/ofa_kernel/` for backward compatibility.

Using the `...-dkms` path directly is more robust as it contains the actual files, avoiding potential issues with symlinks that may be missing, broken, or not yet created during package installation.

### DIFF
--- a/kernel-open/nvidia-peermem/nvidia-peermem.Kbuild
+++ b/kernel-open/nvidia-peermem/nvidia-peermem.Kbuild
@@ -43,17 +43,11 @@ else
     OFA_ARCH := $(ARCH)
 endif
 OFA_DIR := /usr/src/ofa_kernel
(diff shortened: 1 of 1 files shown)


Binary-Eater:
@blochl I believe the original version for specifying the DOCA-OFED kernel path might be better. The main reason I feel this way is I am worried about the scenario where after installing a DKMS kernel, the user decides to install the non-DKMS kernel without cleaning up the previous DKMS kernel install. I believe that can leave `/usr/src/ofa_kernel-dkms/` residue while `/usr/src/ofa_kernel/` points to the correct source.

blochl:
@Binary-Eater, Thanks for the review!
The DOCA DKMS packages remove the `-dkms` directory on removal (in post-remove). But in any case, I have reworked the commit to be more robust and to use the same logic as the DOCA OFED itself uses.
Plus, I found an issue with x86 that would have manifested with DKMS modules, and fixed it in a separate commit.

Binary-Eater:
Closing since we have an internal review.

PR #1032, opened 2026-02-19
Make kernel parameters world-readable

hoshinolina:
This makes it easier for userspace software to check the status of these settings for diagnostic purposes. There's no good reason for them to be root-only.


aritger:
Hi @hoshinolina.  Thank you for the patch.  I've incorporated that into our code base, and made similar changes to nvidia-modeset.ko and nvidia.ko.  For dumb reasons about how our code is managed, I can't easily merge this directly into github.  But, the change will show up in a subsequent driver release.  Thank you!

PR #1038, opened 2026-02-22
kernel-open: fix BTF generation on kernel 6.15+

coleleavitt:
## Summary

On kernel 6.15+, `Makefile.btf` was introduced upstream which sets `PAHOLE` and other BTF-related variables, making the separate assignment in the NVIDIA Makefile redundant and causing build warnings.

This patch uses `$(or ...)` to conditionally set `PAHOLE`, `BTF_PAHOLE_FLAGS_EXTRA`, `HOLEPUNCH`, and `BTF_IDS_FLAGS_EXTRA` only when they are not already defined by the kernel build system.

Split from #1025 per reviewer request — this change is independent of the DisplayPort MST fix.


Binary-Eater:
Hi @coleleavitt,

https://github.com/NVIDIA/open-gpu-kernel-modules/issues/1041 has a more accurate version of this fix related to the changes in kernel 7.0. Will work on top of there.

PR #1040, opened 2026-02-24
Fix NULL pointer dereferences in DisplayPort DSC capability checks

olejaco:
## Summary

This PR fixes two NULL pointer dereference race conditions that cause kernel panics during DisplayPort hotplug disconnection in the DSC (Display Stream Compression) capability checking code.

### 1. `compoundQueryAttachMSTIsDscPossible()` (line 1453)
- Added `dev->parent` NULL guard to the outer conditional before dereferencing
- Falls back to checking the device's own FEC capability when parent is NULL
- Consistent with existing pattern at line 1575

### 2. `compoundQueryAttachMSTDsc()` (line 1551)
- Added NULL check for `dev->devDoingDscDecompression` before calling `populateDscCaps()`
- Protects against teardown race where the pointer could be NULL between the capability check and function call

## Root Cause

During DisplayPort hotplug disconnection, a race condition occurs:
1. Parent device destructor sets `children[i]->parent = 0` (dp_deviceimpl.cpp:72)
2. Child device's `devDoingDscDecompression` may still reference the parent
3. `discoveryLostDevice()` sets `devDoingDscDecompression = NULL` (dp_connectorimpl.cpp:903)
4. If DSC capability queries execute during this teardown window, NULL dereferences occur

## Testing

- Tested on NVIDIA driver 590.48.01 with DKMS
- DisplayPort hotplug stress testing (multiple connect/disconnect cycles)
- No kernel panics observed after fix
- Display functionality verified (resolution, refresh rate, hotplug detection)

## Impact

- Prevents kernel crashes during DisplayPort hotplug events
- Particularly affects systems with MST (Multi-Stream Transport) topologies
- Improves stability for devices with DSC-capable displays

### DIFF
--- a/src/common/displayport/src/dp_connectorimpl.cpp
+++ b/src/common/displayport/src/dp_connectorimpl.cpp
@@ -1450,9 +1450,10 @@ bool ConnectorImpl::compoundQueryAttachMSTIsDscPossible
     {
         if (dev && dev->isDSCPossible())
         {
-            if ((dev->devDoingDscDecompression != dev) ||
-                ((dev->devDoingDscDecompression == dev) &&
-                (dev->isLogical() && dev->parent)))
+            if (dev->parent &&
+                ((dev->devDoingDscDecompression != dev) ||
+                 ((dev->devDoingDscDecompression == dev) &&
+                 (dev->isLogical() && dev->parent))))
             {
                 //
                 // If DSC decoding is going to happen at sink's parent or
@@ -1546,7 +1547,15 @@ bool ConnectorImpl::compoundQueryAttachMSTDsc(Group * target,
     dpMemZero(&dscInfo, sizeof(DSC_INFO));
     dpMemZero(&warData, sizeof(WAR_DATA));
 
-    // Populate DSC related info for PPS calculations
+    /// Populate DSC related info for PPS calculations
+    if (!dev->devDoingDscDecompression)
+    {
+      //
+      // Device torn down during query - DSC no longer valid.
+      // This can happen during hotplug disconnection race conditions.
+      //
+      return false;
+    }
     populateDscCaps(&dscInfo, dev->devDoingDscDecompression, pDscParams->forcedParams);
 
     // populate modeset related info for PPS calculations


Binary-Eater:
Thanks for the change. We are actually tackling the same problem you encountered in nvbug 5871511.
Before being able to review, would mind sharing with me how many monitors you have connected to the dock? If it's more than one, does having only a single monitor connected to the dock not reproduce the issue on hot unplug of the dock?

I tried reading your bug report capture for this information but unfortunately had the following.

```
*** /proc/driver/nvidia-modeset/dpys
*** ls: -r--r--r-- 1 root root 0 2026-02-06 18:54:24.392145253 +0100 /proc/driver/nvidia-modeset/dpys
deviceId                     : 00
 connector                   : DP-0
  dpy                        : (not connected)
 connector                   : DP-1
  dpy                        : (not connected)
 connector                   : DP-2
  dpy                        : (not connected)
 connector                   : DP-3
  dpy                        : (not connected)
 connector                   : DP-4
  dpy                        : (not connected)
 connector                   : DP-5
  dpy                        : (not connected)

____________________________________________

*** /proc/driver/nvidia-modeset/heads
*** ls: -r--r--r-- 1 root root 0 2026-02-06 18:54:24.392145253 +0100 /proc/driver/nvidia-modeset/heads
deviceId                     : 00
 (not yet initialized)
```

Public forum post: https://forums.developer.nvidia.com/t/kernel-null-pointer-dereference-in-nvidia-modeset-during-thunderbolt-dock-disconnect/359280

olejaco:
Hey @Binary-Eater, thanks for the review. I`ll test your suggestions once I get home from work :+1:

I have 2 external monitors on display port in my setup. I must have run the bug report capture with my laptop disconnected from thunderbolt :thinking:

Binary-Eater:
Hi @olejaco, we actually were able to debug this internally and went with a different approach to the solution. However, we did accredit you internally for your help with this. Thanks a lot!

Here is the fix we implemented if you are interested. It landed in the 580.142 driver.

* https://www.nvidia.com/en-us/drivers/details/265587/
* https://github.com/NVIDIA/open-gpu-kernel-modules/commit/1a5075f82128be3603d4a4657c7b689234306778#diff-7cc6c5263a99c4757a37de2a4f86852f079b4a67801b8be1281ef2dc738c0333R68-R71
* https://github.com/NVIDIA/open-gpu-kernel-modules/commit/1a5075f82128be3603d4a4657c7b689234306778#diff-bdd45f7fbc98ec46a1c67371f070cd0017aa9d885231845d25a6a716681e598aR1434-R1435

Binary-Eater:
Interesting that those links do not work that well in a new session. Just in case.

* https://github.com/NVIDIA/open-gpu-kernel-modules/blob/1a5075f82128be3603d4a4657c7b689234306778/src/common/displayport/src/dp_connectorimpl.cpp#L1434-L1435
* https://github.com/NVIDIA/open-gpu-kernel-modules/blob/1a5075f82128be3603d4a4657c7b689234306778/src/common/displayport/src/dp_deviceimpl.cpp#L68-L71

oxyum:
> Here is the fix we implemented if you are interested. It landed in the 580.142 driver.
> 
Thanks! Finally I can use my new dual-monitor KVM!

PS:
- Lenovo Thinkpad P16G Gen3 with NVIDIA RTX PRO 5000 Blackwell GPU
- NixOS 25.11 with kernel from unstable
- Linux kernel 6.18.17
- nvidia-open 580.142

Binary-Eater:
Thanks for the feedback. Closing out this PR!

Issue #1033, opened 2026-02-19
Kernel panic due to unsafe use of nvkms-kapi.h:struct NvKmsKapiCallbacks


roman-zilka:
### NVIDIA Open GPU Kernel Modules Version

590.48.01 and others

### Please confirm this issue does not happen with the proprietary driver (of the same version). This issue tracker is only for bugs specific to the open kernel driver.

- [x] I confirm that this does not happen with the proprietary driver package.

### Operating System and Version

Gentoo Linux

### Kernel Release

6.18.12-gentoo, custom build

### Please confirm you are running a stable release kernel (e.g. not a -rc). We do not accept bug reports for unreleased kernels.

- [x] I am running on a stable kernel release.

### Hardware: GPU

GeForce RTX 5070 and others

### Describe the bug

In the downstream Gentoo Linux bug https://bugs.gentoo.org/969413 multiple users (incl. myself) have reported freezes / kernel panics during system shutdown with various versions of the GPU driver incl. 590.48.01 and various kernel versions incl. 6.18.12.

We've discovered that when the kernel is built with RANDSTRUCT_FULL or RANDSTRUCT_PERFORMANCE, the GPU driver causes the system to do one of the following during shutdown/reboot, just before the physical power-down:
* nothing unusual
* freeze
* kernel panic

![Image](https://github.com/user-attachments/assets/69c96f65-b72d-4b86-9755-42e1b4fb316a)

Which one of the three behaviors occurs is tied to the kernel build. Removing RANDSTRUCT from the kernel fixes the issue, poweroff is always clean then. The message on the attached screenshot has led me to identify `struct NvKmsKapiCallbacks` declared in `nvkms-kapi.h` as the source of the bug: a function contained in it is called unsafely somewhere and when the contents of the struct become shuffled by RANDSTRUCT, a wrong function is called instead. In the screenshot, it is the suspend/resume function, which is out of place on my system - I don't use suspend. When a freeze happens instead of a panic, it is the probe function, presumably.

The following patch disables the randomization of the said struct and ensures the freeze/panic doesn't happen with RANDSTRUCT enabled, confirming the cause of the bug. The effect of the patch has been verified by me (590.48.01 open driver, 6.18.12, RTX 5070) and one other Gentoo user so far (https://bugs.gentoo.org/969413#c41, 580.126.09 proprietary driver, unknown GPU, unknown kernel). It cannot be considered a fix, of course, just a workaround.

```diff
[...]

PR #1042, opened 2026-02-27
*/nvkms-kapi.h: Exclude struct NvKmsKapiCallbacks from RANDSTRUCT

roman-zilka:
#1033

When the RANDSTRUCT plugin (`CONFIG_RANDSTRUCT*`) is applied, the physical layout of `struct NvKmsKapiCallbacks` in `nvidia-drm.ko` and `nvidia-modeset.ko` may become different at compile-time, leading to a freeze or kernel panic at shutdown. Disable randomization for that struct.

### DIFF
--- a/kernel-open/common/inc/nvkms-kapi.h
+++ b/kernel-open/common/inc/nvkms-kapi.h
@@ -604,7 +604,7 @@ struct NvKmsKapiCallbacks {
     void (*remove)(NvU32 gpuId);
     void (*probe)(const struct NvKmsKapiGpuInfo *gpu_info);
 /* https://github.com/NVIDIA/open-gpu-kernel-modules/issues/1033 */
-#if defined(CONFIG_RANDSTRUCT_PERFORMANCE) || defined(CONFIG_RANDSTRUCT_FULL)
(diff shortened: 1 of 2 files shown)


roman-zilka:
The canonical way to disable the compiler attribute is `__no_randomize_layout` from ` <linux/compiler_types.h>`, but the `#include` was avoided to prevent namespace pollution.

roman-zilka:
4.x and 5.x kernels not tested, but that is the name of the config option. The attribute has the same name.

roman-zilka:
Ideally, the following would be put into a common-type header, but I couldn't find a good one. `nv-linux.h` looks suitable, but it's not reachable by all users of `struct NvKmsKapiCallbacks`.

```c
#if defined(CONFIG_RANDSTRUCT_PERFORMANCE) || defined(CONFIG_RANDSTRUCT_FULL) || defined(CONFIG_GCC_PLUGIN_RANDSTRUCT)
#define __nv_no_randomize_layout __attribute__((no_randomize_layout))
#else
#define __nv_no_randomize_layout
#endif
```

Binary-Eater:
Hi, we authored an internal version of this change but made sure to credit you. I will let you know when the change makes it on this repository @roman-zilka

Binary-Eater:
Should be fixed in driver 580.142. The fix should propagate to our future driver releases as well.

roman-zilka:
> Should be fixed in driver 580.142. The fix should propagate to our future driver releases as well.

580.142 seems to work for me. Built four kernels, all shut down cleanly. Thanks!

PR #1068, opened 2026-03-19
Random small fixes in preparation for full kbuild support

minipli-oss:
Various small fixes in preparation for building all sources with Linux's kbuild with the ultimate goal of supporting features like RANDSTRUCT, kCFI and grsecurity kernel features like RAP, KERNEXEC, CONSTIFY,...

PR #1069, opened 2026-03-19
Full kbuild support (atop of https://github.com/minipli-oss/open-gpu-kernel-modules/tree/fixes)

minipli-oss:
This set of changes makes it possible to compile the OS-agnostic part below `src/` to be compiled with Linux's kbuild which brings a lot of benefits, as in automatically picking up the right compiler flags and, hence, code instrumentation to support features like RANDSTRUCT, kCFI and various grsecurity features.

The changes build atop of the `fixes` branch (https://github.com/NVIDIA/open-gpu-kernel-modules/pull/1068).

Most significant changes are to get rid of system headers completely, which was kinda challenging for the C++ parts of `src/nvidia-modeset/`, as well as actually supporting C++ source files via kbuild. I took the poor man's approach and bend the rules for C to get all `*FLAGS*` / `*flags*` handling for free instead of reimplementing it. A "proper" implementation may do something more sophisticated.

Another challenge was to work around a limitation of kbuild for modules with lots of object files. kbuild simply doesn't allow compound object files for modules and there is, apparently, [no interest](https://lore.kernel.org/linux-kbuild/CAK7LNARLiLzqpMfmvVMJ8pT7-wYjfrP54mey_MCMbULLFY_H+A@mail.gmail.com/) in fixing this. So I had to implement a hack similar to [this](https://lore.kernel.org/linux-kbuild/6837671.QJadu78ljV@zen.local/) using a fake module. A little ugly under the hood, but works quite well.

The final obstacle was C++'s COMDAT sections confusing objtool and the linker. After finally understanding how they work, I was able fix that by simply collapsing them prior to linking the final module.

All in all quite some pragmatic choices, therefore just a proposed as a draft. However, I was able to successfully build the modules and actually use them.

Building with kbuild also allows to enable a lot of debugging options. For example, one of my build configs had UBSAN enabled which triggered a few interesting splats:

```
root@gpu-dev:~# dmesg | grep UBSAN -A1
[    8.831486] UBSAN: array-index-out-of-bounds in /home/minipli/src/open-gpu-kernel-modules/kernel-open/../src/nvidia/src/kernel/mem_mgr/io_vaspace.c:444:36
[    8.831490] index 1 is out of range for type 'RmPhysAddr [1]'
--
[    9.119483] UBSAN: array-index-out-of-bounds in /home/minipli/src/open-gpu-kernel-modules/kernel-open/../src/nvidia/arch/nvalloc/unix/src/registry.c:564:60
[    9.119487] index 1 is out of range for type 'PACKED_REGISTRY_ENTRY [*]'
--
[    9.924916] UBSAN: array-index-out-of-bounds in /home/minipli/src/open-gpu-kernel-modules/kernel-open/../src/nvidia/src/kernel/gpu/fifo/arch/maxwell/kernel_fifo_gm107.c:776:45
[    9.924920] index 15 is out of range for type 'NvU32 [15]'
--
[    9.957044] UBSAN: array-index-out-of-bounds in /home/minipli/src/open-gpu-kernel-modules/kernel-open/../src/nvidia/src/kernel/gpu/mem_mgr/mem_desc.c:2496:28
[    9.957047] index 1 is out of range for type 'RmPhysAddr [1]'
--
[    9.969087] UBSAN: array-index-out-of-bounds in /home/minipli/src/open-gpu-kernel-modules/kernel-open/../src/nvidia/src/kernel/mem_mgr/io_vaspace.c:336:68
[    9.969090] index 1 is out of range for type 'RmPhysAddr [1]'
--
[    9.981216] UBSAN: array-index-out-of-bounds in /home/minipli/src/open-gpu-kernel-modules/kernel-open/../src/nvidia/src/kernel/mem_mgr/io_vaspace.c:336:35
[    9.981219] index 1 is out of range for type 'RmPhysAddr [1]'
--
[   10.139401] UBSAN: array-index-out-of-bounds in /home/minipli/src/open-gpu-kernel-modules/kernel-open/../src/nvidia/src/kernel/rmapi/rmapi_cache_handlers.c:107:91
[   10.139470] index 4294967295 is out of range for type '<unknown> [32]'
--
[   10.150784] UBSAN: array-index-out-of-bounds in /home/minipli/src/open-gpu-kernel-modules/kernel-open/../src/nvidia/src/kernel/rmapi/rmapi_cache_handlers.c:108:91
[   10.150787] index 4294967295 is out of range for type '<unknown> [32]'
--
[   12.455149] UBSAN: array-index-out-of-bounds in /home/minipli/src/open-gpu-kernel-modules/kernel-open/../src/nvidia/src/kernel/gpu/mem_mgr/mem_desc.c:2503:32
[   12.455152] index 1 is out of range for type 'RmPhysAddr [1]'
```

Lots of these are false positives, as the code is making use of old-style flexibe arrays. However, the reports for kernel_fifo_gm107.c (letting though an invalid index (15)) and rmapi_cache_handlers.c or (index -1) may deserve a closer look.


minipli-oss:
I just realized, I completely forget about mentioning how to actually build the modules with kbuild:

```shell-session
$ make modules -j $(nproc) USE_KBUILD=1
[...]
$ sudo make modules_install -j $(nproc) USE_KBUILD=1
```

The key is to mention `USE_KBUILD=1` for both, building and installing the modules.

minipli-oss:
I rebased the branch.

The unconditional use of C11's `_Atomic(...)` type wrappers required me including a gcc header, see https://github.com/NVIDIA/open-gpu-kernel-modules/pull/1069/changes/8ca4dca015c0e49f7540cd08e030a726538a1ad6.

I also wrapped the RANDSTRUCT hack for `struct NvKmsKapiCallbacks` as it's not needed for pure kbuild builds, see https://github.com/NVIDIA/open-gpu-kernel-modules/pull/1069/commits/59bc0e11500d12cb6331a67dc872b277750a15a0.

minipli-oss:
Rebased against 610.43.02.

src/common/unix/xzminidec/interface/xz.h
```diff
@@ -257,7 +257,11 @@ XZ_EXTERN void xz_dec_end(struct xz_dec *s);
  */
 #ifndef XZ_INTERNAL_CRC32
 #	ifdef __KERNEL__
-#		define XZ_INTERNAL_CRC32 0
+		/*
+		 * The in-kernel xz_crc32_init() and xz_crc32() functions aren't
+		 * exported for modules, so we have to use our own either way.
+		 */
+#		define XZ_INTERNAL_CRC32 1
```
lucasdemarchi: Isn't this doing rather the opposite what it should? We actually want to use the optimized crc32 implementation. Looking in the kernel source in lib/decompress_unxz.c, the call to the init function is ifdef'ed out if XZ_INTERNAL_CRC32 is 0 and it seems more correct that way: when using a standalone environment (e.g. when bootloader is uncompressing the kernel image or when using this as userspace code) then we need a working crc32 implementation since there's none available.

When this code is built in the kernel or as kernel module, there shouldn't be calls to xz_crc32_init() or xz_crc32(). The _init() could be suppressed by a similar ifdef and the xz_crc32() is already redirected by:

```
#if defined(__KERNEL__) && !XZ_INTERNAL_CRC32
#       include <linux/crc32.h>
#       undef crc32
#       define xz_crc32(buf, size, crc) \
                (~crc32_le(~(uint32_t)(crc), buf, size))
#endif
```

minipli-oss: I tried and even went as far as making the symbols in src/common/unix/xzminidec/src/xz_crc32.c conditional on XZ_INTERNAL_CRC32. However, that lead to linker errors because src/common/unix/nvidia-3d/src/nvidia-3d-fermi.c unconditionally makes use of xz_crc32_init().

Maybe the fix is to make that depend on XZ_INTERNAL_CRC32, I'll try that, thanks!

minipli-oss: Yeah, that seems to work. Thanks for the suggestion!

PR #1070, opened 2026-03-19
Compatibility with grsecurity kernels

minipli-oss:
Building atop of the kbuild changes (https://github.com/NVIDIA/open-gpu-kernel-modules/pull/1069), this branch contains changes required for grsecurity compatibility. It adds compatibility for KERNEXEC, CONSTIFY, PRIVATE_KSTACKS, RESPECTRE, AUTOSLAB and SIZE_OVERFLOW.

Some of the features need to be disabled when compiling C++ code. However, it either doesn't matter (AUTOSLAB wouldn't instrument anything as all allocations get mapped to a single C-wrapper function anyway) or can be handled by a post-processing step in objtool (adding missing `.return_sites` section and alike).

As grsecurity makes use of ASM alternatives also for its compiler plugin based instrumentation, `--gc-sections` needs to be dropped (and as a consequence `-ffunction-sections -fdata-sections` as well), as otherwise critical sections would be dropped as well (`.alt_instructions`, `.return_sites`,...). This leads to bigger modules but seems to have no other negative effect beside the "neat symbol localization trick" for `memcpy()` and `memset()` no longer working, causing endless loops in `os_mem_set()` calling `memset()` from gcc_helper.c calling `os_mem_set()` calling…. I fixed that by simply not linking gcc_helper.o at all and leaving these symbols unresolved, ready for the module loader to pick them up.

The biggest change, however, is making the code base compatible with RAP, grsecurity's implementation of a CFI scheme. RAP is quite picky when it comes to function (pointer) signatures and mismatches of such either during function pointer assignments or calls at runtime. Lots of the function signature mismatches are in code below `src/nvidia/generated/` which seems to be, well, generated code. Patching it by hand makes no sense, so I implemented support for mangling sources prior to compilation using [coccinelle](https://coccinelle.gitlabpages.inria.fr/website/) to apply semantic patches that will fix these (commits https://github.com/NVIDIA/open-gpu-kernel-modules/commit/5f55948cb3ada859ef0bf3fcec7ebfb65d2b6d3c and following up to https://github.com/NVIDIA/open-gpu-kernel-modules/commit/308485fa3f4939183b1621858df0fed2a3cafe21 as yes, the pFunc mangling patches aren't ready yet :/). They're incomplete and probably shouldn't be integrated anyway as the better solution would be generating clean code from the beginning instead of trying to retroactively mangle it further. That, however, is a change you guys have to do on your side.

Finally, https://github.com/NVIDIA/open-gpu-kernel-modules/pull/1070/changes/6be9dae12518bb08a1fa8d05977bf8090cfd8b3c is just a note about a memory leak we noticed via AUTOSLAB. But as this is referring to some internal bug 4292180, I didn't "fix" it.

Therefore this is just an RFC, how to get changes in or workflows adapted on your side.


minipli-oss:
I just figured the pFunc thunk generation and updated the branch. It was just a few more hacks, but seems to be complete now.

I also sped-up the cocci patch generation by making use of some more cores if `make` was invoked with `-j …` and ignoring the `g_bindata*.c` files which don't need patching but eat up a lot of time in cocci.

The generated patches accumulate to something ~3MB, so definitely nothing one would like to do manually.

minipli-oss:
Rebased to current version

minipli-oss:
Rebased again against 610.43.02.

The `pFunc` handling got much easier, as there are only a few cases left (the ones getting only a single argument instead of two) and all the function pointer casts, of course.

I was also able to drop `fix_nvoc_ctor.cocci` as all violations have been taken care of. Thanks!
However, `fix_nvoc_dtor.cocci` still patches two locations.

minipli-oss:
Just for illustration purposes, these are the patches generated by the cocci scripts to fix remaining pFunc issues and missing designated inits, to be applied below `src/nvidia/`: [610.57.04-cocci_patches.tar.gz](https://github.com/user-attachments/files/31072266/610.57.04-cocci_patches.tar.gz)

It's much less since the 610.x.y series, but still enough to keep the coccinelle mangling in the branch.

Issue #1084, opened 2026-03-28
nvidia-drm: force-enabled HDMI connector does not set link-status to Bad when sink stops responding


msedek:
### NVIDIA Open GPU Kernel Modules Version

595.58.03

### Does this happen with the proprietary driver?

Not tested (RTX 5070 Ti / Blackwell only supports open kernel modules)

### Operating System and Version

CachyOS (Arch-based) rolling release

### Kernel Release

Linux msedek-pc 6.19.10-1-cachyos #1 SMP PREEMPT_DYNAMIC x86_64

### Hardware

GPU 0: NVIDIA GeForce RTX 5070 Ti (UUID: GPU-ce08f902-c260-a75d-c651-9713c298c7ea)

### Describe the bug

When an HDMI connector is force-enabled via the kernel parameter `video=HDMI-A-1:e`, the NVIDIA DRM driver's `detect()` callback in `nvidia-drm-connector.c` reads `connector->force` and passes `forceConnected = NV_TRUE` to NvKMS. This correctly keeps the connector status as "connected" regardless of physical state.

However, when the HDMI sink (a Samsung 75" 4K TV in this case) switches to its built-in apps (Netflix, etc.) and stops listening on its HDMI receiver, the TMDS link goes down. The driver never sets the DRM `link-status` property to `DRM_MODE_LINK_STATUS_BAD`. This means userspace compositors (KWin, Mutter, etc.) are never notified that the link degraded, and they never trigger a modeset to re-train it.

When the TV returns to the HDMI input, the display remains blank because no component in the stack knows the link needs re-negotiation.

Per the DRM/KMS specification, `DRM_MODE_LINK_STATUS_BAD` should be set by the driver when link training fails or when the sink disappears, so userspace can force a re-probe. The KWin compositor already has handling for this (forces a modeset when `linkStatus == Bad`), but it never triggers because the driver never sets it.

The relevant code path in `nvidia-drm-connector.c`:
```c
static bool __nv_drm_detect_encoder(...)
{
    switch (connector->force) {
        case DRM_FORCE_ON:
        case DRM_FORCE_ON_DIGITAL:
            pDetectParams->forceConnected = NV_TRUE;
            break;
        ...
    }
    // NvKMS returns connected, link-status is never updated
}
```

### To Reproduce

1. Connect a TV via HDMI to an NVIDIA GPU
2. Add `video=HDMI-A-1:e` to kernel command line (force-enable the connector)
[...]

Issue #1055, opened 2026-03-11
DP-MST monitor stays black after brief signal drop — DRM connector enabled:disabled / dpms:Off while compositor reports active


mnencia:
### NVIDIA Open GPU Kernel Modules Version

595.45.04

### Does this happen with the proprietary driver (of the same version) as well?

Not tested (only open kernel modules available for Blackwell)

### Operating System and Version

Debian 13 (trixie), GNOME 48.7, Wayland

### Kernel Release

6.18.12+deb13-amd64

### Hardware: GPU

NVIDIA RTX PRO 1000 Blackwell Generation Laptop GPU (GB207GLM) + Intel Arc Pro 140T (iGPU, hybrid graphics)

### Describe the bug

External monitors connected via DP-MST through a Dell WD19DCS dock randomly go black and fail to recover. The monitor stays black indefinitely until a VT switch (`sudo chvt 3 && sleep 2 && sudo chvt 2`) forces a full DRM modeset.

**Setup:**
- Dell Pro Max 16 Premium (MA16250), Intel/NVIDIA hybrid graphics
- Dell Performance Dock WD19DCS (USB-C, DP-MST hub)
- Two external 1920x1080 monitors: Dell U2414H (DP-7) and Philips 234E5 (DP-8)
- Internal eDP-1 (1920x1200) on Intel iGPU
- BIOS: Direct Output Mode ON, Hybrid Graphics ON

**What happens:**
The external monitors experience brief random blackouts (1-5 seconds) every 30-60 minutes, which is likely a DP-MST hub signal renegotiation in the dock. Most of the time the monitors self-recover. However, occasionally one monitor fails to recover and stays permanently black.

**DRM state when monitor is black:**
When the monitor is stuck black, Mutter/GNOME still shows it as connected and configured, but the kernel DRM sysfs reveals a state mismatch:

```
# Black monitor (DP-7):
/sys/class/drm/card1/card1-DP-7/status:  connected
/sys/class/drm/card1/card1-DP-7/enabled: disabled
/sys/class/drm/card1/card1-DP-7/dpms:    Off

# Working monitor (DP-8) for comparison:
/sys/class/drm/card1/card1-DP-8/status:  connected
/sys/class/drm/card1/card1-DP-8/enabled: enabled
/sys/class/drm/card1/card1-DP-8/dpms:    On
```

[...]

PR #1085, opened 2026-03-28
nvidia-drm: set link-status to Bad when force-enabled HDMI sink disappears

msedek:
## Summary

When a connector has `DRM_FORCE_ON` (via `video=CONNECTOR:e`), the detect callback passes `forceConnected=NV_TRUE` to NvKMS, which always returns `connected=true`. The DRM `link-status` property is never updated when the physical HDMI sink disappears (e.g. a TV switching to built-in apps like Netflix).

Per the DRM/KMS specification, drivers must set `link-status` to `DRM_MODE_LINK_STATUS_BAD` when the link degrades, so userspace compositors can trigger a modeset for link re-training. KWin already handles `linkStatus == Bad` with `m_forceModeset = true` ([drm_gpu.cpp:332-335](https://invent.kde.org/plasma/kwin/-/blob/master/src/backends/drm/drm_gpu.cpp#L332)), but this path never fires because the driver never sets the property.

## Fix

Add a physical state probe (without `forceConnected`) before the forced detect in `__nv_drm_detect_encoder()`. When the real physical state is disconnected but the connector is force-enabled, set `link-status` to `Bad`. When the sink is physically present, restore it to `Good`.

This is a minimal, self-contained change (24 lines added) in `kernel-open/nvidia-drm/nvidia-drm-connector.c` that enables existing compositor recovery paths without any userspace changes.

## Test setup

- GPU: NVIDIA GeForce RTX 5070 Ti
- Display: Samsung 75" 4K TV via HDMI-A-1 (force-enabled with `video=HDMI-A-1:e`)
- Driver: nvidia-open 595.58.03
- Kernel: 6.19.10 (CachyOS)
- Compositor: KWin 6.6.3 (Wayland)

## Reproduction

1. Connect TV via HDMI, add `video=HDMI-A-1:e` to kernel cmdline
2. Boot, verify TV displays correctly
3. Switch TV to built-in app (Netflix)
4. Switch TV back to HDMI input
5. Display remains blank because link-status stays "Good" and compositor never modesets

## Related

- Issue: #1084
- KWin MR (closed, redirected here per maintainer): https://invent.kde.org/plasma/kwin/-/merge_requests/9034
- Kernel DRM patch (complementary): https://lore.kernel.org/dri-devel/20260328152550.20222-1-miguelsedek@gmail.com/
- Similar issue class: #1055

### DIFF
--- a/kernel-open/nvidia-drm/nvidia-drm-connector.c
+++ b/kernel-open/nvidia-drm/nvidia-drm-connector.c
@@ -100,6 +100,30 @@ __nv_drm_detect_encoder(struct NvKmsKapiDynamicDisplayParams *pDetectParams,
             break;
     }
 
+    /*
+     * For force-enabled connectors, probe the real physical state first
+     * (without forcing) so we can set link-status to Bad when the sink
+     * disappears. This lets userspace compositors trigger a modeset for
+     * link re-training when the sink returns (e.g. a TV switching back
+     * from built-in apps to its HDMI input).
+     */
+    if (pDetectParams->forceConnected) {
+        struct NvKmsKapiDynamicDisplayParams physicalProbe;
+
+        memset(&physicalProbe, 0, sizeof(physicalProbe));
+        physicalProbe.handle = nv_encoder->hDisplay;
+
+        if (nvKms->getDynamicDisplayInfo(nv_dev->pDevice, &physicalProbe)) {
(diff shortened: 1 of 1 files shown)


msedek:
Closing - the root cause was a userspace script preventing DPMS, not a driver bug. The link re-training happens naturally when DPMS cycles correctly. Sorry for the noise.

PR #1088, opened 2026-03-30
fix(uvm): check pRmApi->Control status in nvGpuOpsSetPageDirectory

sdimitro:
nvGpuOpsSetPageDirectory unconditionally dereferences the return value of vaspaceGetPageDirBase() after calling pRmApi->Control() for NV0080_CTRL_CMD_DMA_SET_PAGE_DIRECTORY without checking the status. If the control call fails (e.g. due to GSP RPC resource exhaustion), the error cleanup path in deviceCtrlCmdDmaSetPageDirectory_IMPL calls gvaspaceExternalRootDirRevoke which sets pExternalPDB to NULL. The subsequent vaspaceGetPageDirBase() then returns NULL, and the call to memdescGetPhysAddr() triggers a kernel NULL pointer dereference.

Add the missing status check after pRmApi->Control() and bail out with proper cleanup (re-enabling channels if externally owned, releasing locks, and freeing thread state) on failure.

### DIFF
--- a/src/nvidia/src/kernel/rmapi/nv_gpu_ops.c
+++ b/src/nvidia/src/kernel/rmapi/nv_gpu_ops.c
@@ -8984,6 +8984,16 @@ NV_STATUS nvGpuOpsSetPageDirectory(struct gpuAddressSpace *vaSpace,
                              NV0080_CTRL_CMD_DMA_SET_PAGE_DIRECTORY,
                              &params,
                              sizeof(params));
+    if (status != NV_OK)
+    {
+        if (vaspaceIsExternallyOwned(pVAS))
+        {
+            nvGpuOpsEnableVaSpaceChannels(vaSpace);
+        }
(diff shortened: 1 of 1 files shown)


sdimitro:
The NVONLINE BUG for this is `6029889`

sdimitro:
@aritger @mmaneetsingh - can I get a confirmation either here or in the nvonline bug that someone is picking this patch up for next release and/or LTS? we have a customer at CoreWeave that's hitting this frequently. Doesn't have to be this exact patch if you don't like it - I care more about the issue being resolved.

aritger:
Hi @sdimitro.  Thank you for reporting the bug and proposing a fix.  Yes, this bug will be addressed in a future release.  I can't promise specific release dates or specific release branches.  But, the fix is making its way through the system.  Thank you.

sdimitro:
Closing this. This fix is now part of 580.173.02+

PR #1096, opened 2026-04-03
nvidia: fix mmap_context use-before-initialization in vma access

Romani-Archman:
Assign mmap_context before using mmap_context->access_start in nvidia_vma_access().

### DIFF
--- a/kernel-open/nvidia/nv-mmap.c
+++ b/kernel-open/nvidia/nv-mmap.c
@@ -132,11 +132,6 @@ nvidia_vma_access(
     NvBool bIsNuma = NV_FALSE;
     int ret = -EINVAL;
 
-    bIsNuma = pfn_valid(mmap_context->access_start >> PAGE_SHIFT);
-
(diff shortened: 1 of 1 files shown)


Geremia:
@Romani-Archman  I thought your patch (∵ it involves VMA) might've helped me fix #1113, but it didn't. 🤔

Romani-Archman:
> @Romani-Archman I thought your patch (∵ it involves VMA) might've helped me fix #1113, but it didn't. 🤔

I took a look at the issue you mentioned. The problem you're seeing is likely related to `VMA_LOCK_OFFSET` being renamed in Linux 7.0 kernel. My PR/commit actually addresses a bug where `mmap_context` wasn't being initialized in `nvidia_vma_access()`, which was causing NULL pointer dereference.

sollniss:
Why was this closed? Pretty sure this fixes #1107.

Romani-Archman:
> Why was this closed? Pretty sure this fixes #1107.

595.71.05 fixes this issue in a more proper way, so I have closed this PR.

Issue #860, opened 2025-05-20
Battery ACPI event wakes up dGPU, shortening battery life on Linux


PetrusCrus:
### NVIDIA Open GPU Kernel Modules Version

570.144

### Please confirm this issue does not happen with the proprietary driver (of the same version). This issue tracker is only for bugs specific to the open kernel driver.

- [ ] I confirm that this does not happen with the proprietary driver package.

### Operating System and Version

Description:	NixOS 25.05 (Warbler)

### Kernel Release

Linux Zephyrus 6.14.6 #1-NixOS SMP PREEMPT_DYNAMIC Fri May  9 07:56:10 UTC 2025 x86_64 GNU/Linux

### Please confirm you are running a stable release kernel (e.g. not a -rc). We do not accept bug reports for unreleased kernels.

- [x] I am running on a stable kernel release.

### Hardware: GPU

GPU 0: NVIDIA GeForce RTX 4060 Laptop GPU (UUID: GPU-15764050-a2ea-30c8-e844-a5f96fbe0389)

### Describe the bug

When battery discharges, it will send a ACPI notification that wakes up dGPU on multi-gpu systems, which shortens battery life

### To Reproduce

To reproduce this issue, I leave my laptop idle, and run the following scripts in terminal
```
sudo bpftrace -e '
kretprobe:nv_pmops_runtime_resume
{
    printf("nv_pmops_runtime_resume returned %d for process %s (pid=%d)\n", retval, comm, pid);
    print(kstack(20));
}
'
```
and when my dGPU got waken up, it prints:
```
nv_pmops_runtime_resume returned 0 for process kworker/0:1 (pid=5380)

        __rpm_callback+65
        rpm_callback+89
        rpm_resume+1243
        __pm_runtime_resume+75
        pci_device_shutdown+34
        nv_indicate_not_idle+43
        os_ref_dynamic_power+342
        rm_acpi_nvpcf_notify+131
        acpi_ev_notify_dispatch+70
        acpi_os_execute_deferred+23
        process_one_work+394
        worker_thread+564
        kthread+235
        ret_from_fork+49
        ret_from_fork_asm+26
```
which indicate that some ACPI event wakes it up. So I run `sudo acpi_listen`, and got `battery PNP0C0A:00 00000080 00000001`.

[...]

PR #1181, opened 2026-06-06
Don't wake a runtime-suspended dGPU to service NVPCF/GPS ACPI notifies

ElXreno:
On RTD3 laptops the discrete GPU sits in D3cold while idle. On some machines the platform keeps delivering ACPI `Notify()` events (to the NVPCF device, and as GPS status changes) even while the GPU is suspended, for example around battery/AC transitions or when the SBIOS pushes a new thermal or power-limit hint.

The two handlers that service those notifies, `rm_acpi_nvpcf_notify()` and `RmHandleGPSStatusChange()`, both call `os_ref_dynamic_power()` unconditionally. That resumes the GPU purely to deliver an event that does nothing while it's powered down, and it then re-suspends. On a fair number of laptops the next notify lands right away, so the GPU never settles in D3cold. Folks in #860 describe it cycling D0/D3cold every ~11 seconds on battery, and the current workaround is to patch the ACPI tables to strip the `Notify(NPCF, 0xC0)`.

This skips the resume when the GPU is already runtime-suspended (`NV_DYNAMIC_POWER_STATE_IDLE_INDICATED`), the same guard `rm_pmu_perfmon_get_load()` already uses a few functions away. Why it's safe:

* the NVPCF event is only consumed by nvidia-powerd, which has nothing to do while the GPU is asleep
* GPS/SBIOS data is re-read on the next `StateLoad`, so a skipped sync is recovered when the GPU next powers up
* power-source (AC/battery) changes go through a separate path (`rm_power_source_change_event`) and are left untouched

I traced the wake on an RTX 4060 mobile (ASUS TUF, open module):

```text
acpi_ev_notify_dispatch -> rm_acpi_nvpcf_notify -> os_ref_dynamic_power
  -> nv_indicate_not_idle -> ... -> acpi_pci_set_power_state (D3cold -> D0)
```

With the guard in place the handlers still run, but the GPU stays in D3cold. Verified on 595.45.04 and 610.43.02.

### DIFF
--- a/src/nvidia/arch/nvalloc/unix/src/osapi.c
+++ b/src/nvidia/arch/nvalloc/unix/src/osapi.c
@@ -2023,6 +2023,17 @@ static void RmHandleGPSStatusChange
     RM_API            *pRmApi;
     THREAD_STATE_NODE  threadState;
     NV0000_CTRL_SYSTEM_GPS_CONTROL_PARAMS gpsControl = { 0 };
+    nv_priv_t         *nvp = NV_GET_NV_PRIV(pNv);
+
+    //
+    // Don't wake a runtime-suspended GPU just to handle a GPS status
+    // change; service it only when the GPU is already awake.
+    //
+    if (nvp != NULL &&
+        nvp->dynamic_power.state == NV_DYNAMIC_POWER_STATE_IDLE_INDICATED)
+    {
+        return;
+    }
 
     pRmApi = RmUnixRmApiPrologue(pNv, &threadState, RM_LOCK_MODULES_ACPI);
     if (pRmApi == NULL)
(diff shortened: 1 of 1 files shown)


ryanabx:
Possibly fixes: #905 

There is in fact a `notify` call in the Legion 16APH{8,9} ACPI tables:
https://github.com/NVIDIA/open-gpu-kernel-modules/issues/905#issuecomment-4112182871

Other commenters with different laptops have noticed similar issues. I myself with a 16APH8 have noticed the around 11-15 second consistent sleep/wake when no applications are using the dGPU.

Kelsios:
I have applied the patch to nvidia-open-dkms 610.43.03, on an ASUS Zephyrus G16 (RTX5080), and I can confirm it works. I have also set up a pacman hook, until this is merged into upstream.
OS: Arch Linux, kernel 7.1.5-arch1-2
Thanks a lot @ElXreno !!

felipecwb:
Interesting... I believe this will fix my issue where nvidia awakes periodically on battery.
I might try to complie with this patch and hope to see the this  officially released.

lars76:
Thank you for working on this. I tested the NVPCF part on an ASUS ROG Zephyrus G14 GA403UI with an RTX 4070 Laptop GPU, NVIDIA driver 610.57.04, and Linux 7.1.6.

I confirmed that suppressing NVPCF notifications while the GPU is runtime-suspended prevents the unnecessary wake. On my laptop, the firmware sends one notification per 1% of battery discharge. Without the guard, each notification wakes the GPU for approximately 20–22 seconds.

Each wake consumed 110.4 ± 19.1 mWh. Since one occurs for every 1% discharged, approximately 17% of the observed battery drain was spent on these wakes while the NVIDIA GPU remained otherwise unused.

On the tested laptop, provided the NVIDIA GPU remains otherwise unused, the patch reduces battery consumption by approximately 17%.

However, I found one issue with dropping the notification rather than deferring it. The notification tells nvidia-powerd "the power limits changed, re-read them", and powerd never re-reads on its own. In my test I switched the platform profile while the GPU was suspended with notifications dropped: the GPU stayed capped at the old profile's 75 W instead of the new 90 W, even after it woke up, until the next battery-percent notification arrived.

I implemented a variant that saves the notification and delivers it on the next resume instead: #1299. Same zero wakes on battery, but the limits always end up correct. I left GPS unchanged because I have not tested whether suppressing it is safe.

Issue #1177, opened 2026-06-03
DIFR Prefetch causes a freeze in pop_os


Ilya122:
### NVIDIA Open GPU Kernel Modules Version

NVIDIA Open GPU Kernel Module version: 580.159.03 Full version string: NVIDIA UNIX Open Kernel Module for x86_64 580.159.03 Release Build Build date: Fri Apr 24 06:03:03 UTC 2026 Kernel: 6.18.7-76061807-generic GPU: GeForce RTX 5070

### Please confirm this issue does not happen with the proprietary driver (of the same version). This issue tracker is only for bugs specific to the open kernel driver.

- [ ] I confirm that this does not happen with the proprietary driver package.

### Operating System and Version

Pop!_OS 24.04 LTS

### Kernel Release

Linux pop-os 6.18.7-76061807-generic #202601231045~1778249322~24.04~b44a3c3 SMP PREEMPT_DYNAMIC Fri M x86_64 x86_64 x86_64 GNU/Linux

### Please confirm you are running a stable release kernel (e.g. not a -rc). We do not accept bug reports for unreleased kernels.

- [x] I am running on a stable kernel release.

### Hardware: GPU

GPU 0: NVIDIA GeForce RTX 5070

### Describe the bug

My PC freezes from time to time (currently experienced once in a day)

Journal suggests it's something in nvidir DIFR

LOG:
Jun 03 10:40:15 pop-os kernel: watchdog: BUG: soft lockup - CPU#2 stuck for 104s! [nvidia-modeset/:1314]
Jun 03 10:40:15 pop-os kernel: Modules linked in: r8153_ecm cdc_ether usbnet r8152 mii uhid ccm rfcomm snd_seq_dummy >
Jun 03 10:40:15 pop-os kernel:  snd_compress snd_intel_dspcfg snd_usbmidi_lib btmtk snd_intel_sdw_acpi ac97_bus btrtl>
Jun 03 10:40:15 pop-os kernel:  ghash_clmulni_intel ahci nvme_keyring nvme_auth libahci realtek hkdf intel_vsec video>
Jun 03 10:40:15 pop-os kernel: CPU: 2 UID: 0 PID: 1314 Comm: nvidia-modeset/ Tainted: G           OEL      6.18.7-760>
Jun 03 10:40:15 pop-os kernel: Tainted: [O]=OOT_MODULE, [E]=UNSIGNED_MODULE, [L]=SOFTLOCKUP
[...]

Issue #1167, opened 2026-05-28
Display freezes requiring reboot usually shortly after resuming from suspend


Gert-dev:
### NVIDIA Open GPU Kernel Modules Version

595.71.05

### Please confirm this issue does not happen with the proprietary driver (of the same version). This issue tracker is only for bugs specific to the open kernel driver.

- [ ] I confirm that this does not happen with the proprietary driver package.

### Operating System and Version

Arch Linux

### Kernel Release

7.0.3

### Please confirm you are running a stable release kernel (e.g. not a -rc). We do not accept bug reports for unreleased kernels.

- [x] I am running on a stable kernel release.

### Hardware: GPU

NVIDIA GeForce RTX 4090 Laptop GPU

### Describe the bug

Every so often, usually a couple of minutes after resuming from suspend and unlocking the screen on a GNOME 50.1 system, the display attached to the NVIDIA dGPU will fully freeze.

### To Reproduce

1. Enter suspend.
2. Resume from suspend.
3. Browse the web for a few minutes.

### Bug Incidence

Sometimes

### nvidia-bug-report.log.gz

[nvidia-bug-report.log.gz](https://github.com/user-attachments/files/27275754/nvidia-bug-report.log.gz)

### More Info

I'm running GNOME 50.1 and have a hybrid GPU system where one monitor is attached to the NVIDIA dGPU directly.

Only a forced reboot works when this happens. If music is playing, it will keep on playing, suggesting a freeze in the display stack somewhere. None of these work:
- Dropping to a TTY.
- Replugging the monitor, not even to try and use the internal display.

If SysRQ is enabled, a soft or hard kill don't fix the problem, but they do allow a stacktrace to be logged which seems to point to the NVIDIA driver:

```
19:57:19 computer kernel: sysrq: Show backtrace of all active CPUs
19:57:19 computer kernel: NMI backtrace for cpu 0
19:57:19 computer kernel: CPU: 0 UID: 0 PID: 0 Comm: swapper/0 Tainted: G           OE       7.0.10-arch1-1 #1 PREEMPT(full)  b38726df0ec1c5aec6f05d4eab858505a5944d02
19:57:19 computer kernel: Tainted: [O]=OOT_MODULE, [E]=UNSIGNED_MODULE
19:57:19 computer kernel: Hardware name: LENOVO 82WS/LNVNB161216, BIOS LPCN61WW 03/21/2025
19:57:19 computer kernel: Call Trace:
[...]

PR #1192, opened 2026-06-11
Bound the GPFIFO-space wait in nvWriteGpEntry()

Lumzdas:
### Problem

nvWriteGpEntry() waits for a free GPFIFO entry with no timeout. __nvPushMakeRoom() waits for pushbuffer space with no timeout. Both loops exit only when the GPU makes progress or a channel error is flagged. If a channel stalls without a flagged error, the loop spins forever in kernel context. When the spinning code holds the nvkms lock, every later modeset blocks in D state and the desktop is unusable until reboot.

I hit this reproducibly with DIFR prefetch after resume from suspend. Same trace as #1167 and #1177.

Environment: RTX 5090, driver 595.71.05 open modules, NVreg_PreserveVideoMemoryAllocations enabled, kernel 7.0.3, KDE Plasma on Wayland, S3 suspend.

The nvidia-modeset kthread spins on one core:

```
RIP: 0010:nvWriteGpEntry+0x90/0x370 [nvidia_modeset]
Call Trace:
 nvPushKickoff+0x24/0x40 [nvidia_modeset]
 PrefetchHelperSurfaceEvo+0x5aa/0x610 [nvidia_modeset]
 nvDIFRPrefetchSurfaces+0xed/0x240 [nvidia_modeset]
 DifrPrefetchEventDeferredWork+0x16/0x30 [nvidia_modeset]
```

and the compositor blocks on the nvkms lock held by that thread:

```
task:kwin_wayland state:D
 down+0x45/0x50
 nvkms_ioctl_from_kapi_try_pmlock+0x36/0x80 [nvidia_modeset]
 ApplyModeSetConfig+0xc7a/0xe00 [nvidia_modeset]
 nv_drm_atomic_check+0x220/0x270 [nvidia_drm]
```

### Fix

Commit 1 bounds nvWriteGpEntry() the same way IdleChannel() already does: poll nvPushImportGetMilliSeconds(), honor the channel noTimeout flag, log an error and return FALSE after NV_PUSH_NOTIFIER_SHORT_TIMEOUT. Kickoff() already handles the FALSE return.

Commit 2 bounds __nvPushMakeRoom() with the same timeout. This function returns void and callers write methods unconditionally after it returns, so on timeout it wraps the write pointer back to the buffer base and marks the whole buffer free. Each method write is asserted smaller than the buffer, so the writes stay in bounds, and they land only in the channel's own pushbuffer. The following kickoff then fails and the caller gives up. Commit 2 is needed because with only commit 1 applied the same stall spins in __nvPushMakeRoom() instead, one call earlier on the same DIFR path. Observed on my machine.

DIFR prefetch already has a failure path for a kickoff that never completes: the semaphore wait in PrefetchSingleSurface() times out and reports NV2080_CTRL_LPWR_DIFR_PREFETCH_FAIL_CE_HW_ERROR. That path was unreachable only because the waits above it could spin first.

Channels created with noTimeout = TRUE keep the old behavior.

### Results

Both commits, rebased onto 595.71.05, have been running on my machine since June 25. The stall occurred six times since then, mostly within seconds of resume. Each time the desktop kept working and the driver logged:

```
nvidia-modeset: ERROR: GPU:0: Timed out waiting for a free GPFIFO entry.
nvidia-modeset: ERROR: GPU:0: Timed out waiting for room in the pushbuffer.
```

### DIFF
--- a/src/common/unix/nvidia-push/src/nvidia-push.c
+++ b/src/common/unix/nvidia-push/src/nvidia-push.c
@@ -627,6 +627,7 @@ void __nvPushMakeRoom(NvPushChannelPtr push_buffer, NvU32 count)
     NvU32 getOffset;
     NvU32 putOffset;
     NvBool fenceToEnd = FALSE;
+    NvU64 baseTime, currentTime;
 
     putOffset = (NvU32) ((char *)push_buffer->main.buffer -
                          (char *)push_buffer->main.base);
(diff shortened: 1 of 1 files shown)


Vedaru:
Also hitting this on 595.84 with an RTX 5060 Laptop (Blackwell, GB206M,device 10de:2d59). Ubuntu 26.04, GNOME Wayland, custom kernel 7.1.3.

After a few hours idle at the lock screen, gnome-shell memory climbs from~500 MB to 700+ MB, DIFR prefetch cycles accumulate, and eventually the shell crashes with a SIGSEGV in the cursor/KMS path. Journal shows the compositor falling back to non-atomic mode setting for nvidia-drm, and the KMS thread blocks. Only a hard reboot recovers.

Tried building from this PR's branch, but the GitHub source diverges from Ubuntu's DKMS tarball enough that the resulting modules don't load properly on my system. Would really appreciate this getting merged so it ships in a future driver update. Happy to provide logs if needed

ThatSava:
Also hitting this problem on an Asus Zephyrus G14 with RTX4070. I haven't yet verified this fix, but my investigation points to the same part of the code.
I also have logs that I am willing to provide if needed to fix this and hope it will be swiftly merged and shipped in a future update.
Edit: Also fixes #1246

PR #1186, opened 2026-06-09
modeset: honor forced FRL rate during link assessment

Rohithmatham12:
## Summary
- Treat `nvidia-modeset.force_frl_rate` as an explicit opt-in to use EDID-declared FRL capabilities during HDMI FRL link assessment.
- Leave the default physical link-training assessment path unchanged when `force_frl_rate=0`.
- Update the module parameter description so `modinfo` reflects the EDID-capability behavior.

This helps forced/headless HDMI connectors where a valid override EDID advertises FRL (for example Max_FRL_Rate=6) but physical link assessment cannot complete because there is no real sink handshake, causing high-bandwidth modes such as 4K120 to be pruned.

Fixes #1184.

## Testing
- `git diff --check`
- `make -C kernel-open -n modules`
  - Reached the kernel build handoff and then stopped in this macOS workspace because `/lib/modules/25.5.0/build` is not present.
- Reporter validation on #1184 / this PR:
  - Hardware/software: RTX 3060 Ti, `nvidia-open 610.43.02`, kernel `7.0.11`
  - Boot setup: `drm.edid_firmware=HDMI-A-1:edid/<file> video=HDMI-A-1:e`
  - Override EDID advertises `Max_FRL_Rate=6` plus VIC 118
  - Stock driver with `nvidia-modeset.force_frl_rate=1` and `=2` still stayed capped at 4K60
  - With this patch plus `nvidia-modeset.force_frl_rate=1`, `3840x2160@119.88` appears and sets on the force-enabled HDMI connector
  - HDR also enables and Sunshine HEVC Main10 streaming was confirmed end to end
  - Custom high-bandwidth DisplayID timings such as `3024x1890@120` and `2752x2064@120` also validate and set

Issue #1184, opened 2026-06-07
Force-enabled virtual connector won't bring up FRL — capped at 4K60 despite EDID Max_FRL_Rate=6 (blocks headless 4K120 Sunshine/Moonlight)


HarryAnkers:
Hey I've been relentlessly trying to get 4k 120hz to work on sunshine / moonlight with the latest version of cachyos. I've tried overything but buying a 4k120hz monitor to stub the monitor. All virtual monitor solutions seem borked on the nvidia side sadly. Please may someone look at this. Heres my 3 hours of clauding summary:

Setup: RTX 3060 Ti on nvidia-open 610.43.02, CachyOS, KDE Plasma 6.6 on Wayland, kernel 7.x. The goal is a headless virtual display so I can stream 4K120 HDR to a Mac over Moonlight with nothing physically plugged into the GPU.

What works: I can force a virtual display with `drm.edid_firmware` + `video=HDMI-A-1:e` and a custom EDID, and SDR up to 4K60 is fine. I even got HDR working, but only on the HDMI connector and only with a real, well-formed EDID (a dumped AOC monitor EDID). HDR enable gets rejected on DisplayPort, and it's also rejected with a hand-built EDID even on HDMI. So the driver does honor HDR signaling as long as the sink descriptor looks real enough.

What doesn't work: 4K120. No matter what I do the forced connector caps at 4K60. The reason seems to be that 4K120 needs FRL (HDMI 2.1) or DSC, and the driver won't bring either up unless there's a real sink doing the SCDC/DPCD handshake.

The bit that makes me think this is a driver policy rather than a real limit: I fed it an EDID that declares Max_FRL_Rate=6 (48 Gbps) with VIC 118 (3840x2160@120) in the normal RGB video data block. edid-decode validates it as clean, and the kernel reads it back fine:

    cat /sys/class/drm/card1-HDMI-A-1/edid | edid-decode
    ...
    Max Fixed Rate Link: 3 and 6 Gbps per lane on 3 lanes, 6, 8, 10 and 12 Gbps on 4 lanes

Despite that, the driver only ever offers 3840x2160@50/59/60 and prunes 4K120. So it isn't the EDID being malformed; the driver is reading the FRL declaration and still refusing to bring up the high-bandwidth link because nothing is physically attached.

[...]

PR #1195, opened 2026-06-11
dp: assume sink capabilities on force-enabled connectors without DPCD

HarryAnkers:
Hey I've been trying to get virtual displays working for sunshine game streaming. I really want 4k 120 to work for my MacBook Pro to stream my display. To do this and still have my tv connected I this to work on display port.
Obviously this is mostly made with the help of Claude but is tested and works. Please let's get this merged in so I can move to cachy os haha

Hers the Claude description and shout out to inspired pr and issue — #1186 (thanks @Rohithmatham12, this PR follows your approach on the DP side) and #1184:

---

## Problem

A DisplayPort connector that is force-enabled with an EDID override (`video=DP-n:e` + `drm.edid_firmware=...`) and has no physical sink attached is capped at modes that fit the bare assumed link bandwidth — in practice 4K60 — and HDR cannot be enabled on it at all. This makes DP unusable for headless / virtual-display setups (Sunshine/Moonlight game streaming hosts and similar), where the framebuffer is consumed by a screen-capture encoder rather than a physical monitor. This is the DP side of the limitation described in #1184; #1186 addresses the HDMI FRL side.

## Root cause

With no physical sink there is no HPD, so `nvDpyGetDynamicData()` short-circuits the force-connected case without ever notifying the DP library — no discovery runs and no device is created. Even if a long pulse is delivered, `parseAndReadCaps()` fails its AUX read and leaves `revisionMajor = 0`, which makes `notifyLongPulseInternal()` bail out before discovery (`if (!hal->isAtLeastVersion(1, 0)) goto completed`). The result is that mode validation has no sink DSC/FEC caps (so anything above the bare link rate is pruned) and `nvDpyIsHDRCapable()` fails its DPCD revision check (so HDR enable is rejected).

## Change

Four small steps, all gated on the existing `dpcdOffline` state so connectors with a responsive sink take exactly the same paths as before:

1. **`nvDpyGetDynamicData()`** — when a dpy is force-connected on a DP connector the DP library considers unplugged, simulate a long pulse so discovery runs and a device is created.
2. **`DPCDHALImpl::parseAndReadCaps()`** — on a failed caps read, fall back to `populateFakeDpcd()` (previously only done on the unplug path) so discovery can proceed. The faked DPCD revision is raised from 1.1 to 1.4 since HDR requires DPRX 1.3+.
3. **`DeviceImpl::getDSCSupport()` / `readAndParseDSCCaps()` / `getFECSupport()`** — when the AUX read fails and DPCD is offline, synthesize DSC decoder caps modeled on a typical DSC 1.2 4K high-refresh monitor, and claim FEC, so high-bandwidth modes validate through the DSC path.
4. **`DPCDHALImpl::getSDPExtnForColorimetry()`** — claim VSC SDP colorimetry support when DPCD is offline so HDR signaling can be enabled.

## Testing

RTX 3060 Ti (GA104), driver 610.43.02, kernel 7.0.11, KDE Plasma 6.6 Wayland, connector force-enabled via `drm.edid_firmware=DP-2:edid/<file> video=DP-2:e` with an EDID declaring 3840x2160@120:

- Before: connector offers 3840x2160 up to 60 Hz only; HDR enable is rejected by the driver.
- After: discovery runs (`DP> HPD v1.4`, `DPCONN> New device | Native DSC Capability - Capable`), 3840x2160@120 validates and **sets** (DSC Mode = SINGLE), and HDR + wide color gamut enable end to end. Verified streaming the result with Sunshine (KMS capture, HEVC Main10).
- A physically connected DP monitor on the same GPU continues to behave normally (its DPCD reads succeed, so none of the new paths are taken).

Expected, benign noise when the mode is set: `DP-DEV> Error querying DSC Enable State!` — the post-modeset confirmation read has no sink to talk to.

## Open questions for review

- `parseAndReadCaps()` now falls back to fake caps for *any* failed caps read, not just user-forced connectors — the DP library has no visibility into the force-connected flag without new plumbing. The AUX read already retries 16 times before failing, and the unplug path has populated fake DPCD this way for a long time, but if you'd prefer this to be opt-in (a regkey like `APPLY_MAX_LINK_RATE_OVERRIDES`, or plumbing the forced flag through `notifyLongPulse`), happy to rework it.
- The fake DPCD revision bump (1.1 → 1.4) also applies to the existing EDID-locked fake-device path. I didn't find anything that depends on it staying 1.1, but flagging it since that path predates this change.
- The synthesized DSC cap values are my best reading of `parseDscCaps()` and the DPCD 1.4 spec; if there's a preferred canonical set for a virtual sink, glad to switch to it.

Tested in combination with #1186 (the two are independent — this touches only the DP path). Happy to rework any of the above based on feedback.

### DIFF
--- a/src/common/displayport/src/dp_configcaps.cpp
+++ b/src/common/displayport/src/dp_configcaps.cpp
@@ -185,6 +185,12 @@ void DPCDHALImpl::parseAndReadCaps()
         // Set an invalid state here and make sure we REMEMBER we couldn't get the caps
         caps.revisionMajor = 0;
         dpcdOffline = true;
+
+        //
+        // Forced/sink-less connector (EDID override, no AUX): fall back to
+        // fake caps so discovery can proceed and create a device.
+        //
+        populateFakeDpcd();
         return;
     }
 
@@ -634,6 +640,14 @@ bool DPCDHALImpl::getSDPExtnForColorimetry()
 {
     bool bSDPExtnForColorimetry = false;
     NvU8 byte = 0;
+    if (dpcdOffline)
+    {
+        //
+        // Forced/sink-less connector: claim VSC SDP colorimetry support so
+        // HDR signaling can be enabled. No real sink interprets the SDP.
+        //
+        return true;
+    }
     if (caps.extendedRxCapsPresent)
     {
         if (AuxRetry::ack == bus.read(NV_DPCD14_EXTENDED_DPRX_FEATURE_ENUM_LIST, &byte,  sizeof byte))
@@ -996,7 +1010,7 @@ void DPCDHALImpl::populateFakeDpcd()
     // this should be extended in for more dpcd offsets in future.
     //
     caps.revisionMajor = 0x1;
-    caps.revisionMinor = 0x1;
+    caps.revisionMinor = 0x4;
     caps.supportsESI = false;
     caps.maxLinkRate = dp2LinkRate_8_10Gbps;
     caps.maxLaneCount = 4;
--- a/src/common/displayport/src/dp_deviceimpl.cpp
+++ b/src/common/displayport/src/dp_deviceimpl.cpp
@@ -1732,6 +1732,15 @@ NvBool DeviceImpl::getDSCSupport()
         }
     }
 
+    else if (hal->isDpcdOffline())
+    {
+        //
+        // Forced/sink-less connector (EDID override, no AUX): assume the
+        // virtual sink can decompress DSC so that high-bandwidth modes pass
+        // validation. The stream is never decoded by a real device.
+        //
+        dscCaps.bDSCDecompressionSupported = true;
+    }
     else
     {
         DP_PRINTF(DP_ERROR, "DP-DEV> DSC Support AUX READ failed for %s!", address.toString(sb));
@@ -2308,6 +2317,11 @@ bool DeviceImpl::getFECSupport()
         bFECSupported = this->bandwidth.enum_path.bPathFECCapable;
     }
 
+    else if (hal->isDpcdOffline())
+    {
+        // Forced/sink-less connector: claim FEC so the DSC path is usable.
+        bFECSupported = true;
+    }
     else if (AuxBus::success == this->getDpcdData(NV_DPCD14_FEC_CAPABILITY,
         &byte, sizeof(byte), &size, &nakReason))
     {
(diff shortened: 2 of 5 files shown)


Rohithmatham12:
Thanks for carrying the DP side forward and for testing it on the same headless streaming setup. The problem split makes sense to me: #1186 handles HDMI FRL, while this needs to make the DP DSC/HDR path usable when a connector is force-enabled but has no physical AUX/DPCD sink.

One implementation detail I would tighten before maintainer review is the scope of the fake capability fallback. The current patch makes `DPCDHALImpl::parseAndReadCaps()` call `populateFakeDpcd()` on any failed caps read. That fixes the force-enabled virtual connector case, but it also changes behavior for physical DP sinks with transient or broken AUX reads.

A safer shape would be to make the synthesized DPCD/DSC/FEC/HDR behavior explicitly virtual-sink scoped, for example:

- add a DP-library-visible flag for "forced/offline virtual sink" when `nvDpyGetDynamicData()` handles `pRequest->forceConnected` on a DPLib connector with no plugged device;
- only allow the `parseAndReadCaps()` fake-DPCD fallback and the synthetic DSC/FEC/VSC colorimetry paths when that flag is set;
- clear the flag on a real unplug/plug transition or when DPCD reads succeed, so physical monitors with responsive AUX keep the existing behavior.

That would keep the good tested result here while matching the risk profile of #1186: default behavior unchanged, physical sink behavior unchanged, and virtual/headless behavior enabled only when the user has explicitly force-enabled the connector with an EDID override.

The reporter validation is strong. I think the main question for NVIDIA review will be whether the fake DPCD/DSC/HDR capability path is narrowly gated enough.

Rohithmatham12:
I put together a narrower variant of this idea in #1196 that tries to address the main scoping concern: fake DPCD/DSC/FEC/HDR capabilities are only enabled when NVKMS is handling a force-connected DPLib connector with no plugged DP device.

So instead of falling back to fake DPCD caps on any failed caps read, it threads a virtual-sink fallback flag from the `video=DP-*:e` / EDID-override path into DPLib and gates the synthesized capabilities on that flag.

If you have time to test it on the same DP setup, the interesting result would be whether #1196 still exposes/sets 4K120 + HDR like this PR does, while keeping the behavior more narrowly scoped for maintainers.

Issue #1205, opened 2026-06-16
[595.71.05 open] CPU soft lockup in nvDIFRPrefetchSurfaces on S3 resume — Blackwell mobile (10de:2c59), kernel 6.17


cpfarher:
### NVIDIA Open GPU Kernel Modules Version

595.71.05

### Please confirm this issue does not happen with the proprietary driver (of the same version). This issue tracker is only for bugs specific to the open kernel driver.

- [x] I confirm that this does not happen with the proprietary driver package.

### Operating System and Version

Ubuntu 24.04.4 LTS

### Kernel Release

uname -a Linux christian-Legion-Pro-7-16IAX10H 6.17.0-35-generic #35~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 19:30:42 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

### Please confirm you are running a stable release kernel (e.g. not a -rc). We do not accept bug reports for unreleased kernels.

- [x] I am running on a stable kernel release.

### Hardware: GPU

NVIDIA open kernel module 595.71.05

### Describe the bug

On resume from S3 suspend, a nvidia-modeset kthread soft-locks one CPU indefinitely inside the DIFR prefetch path (nvDIFRPrefetchSurfaces / DifrPrefetchEventDeferredWork); the display never recovers and the machine must be hard-powered-off. 
Reproduces across multiple suspend cycles. Full trace and nvidia-bug-report.log attached.

### To Reproduce

Suspend the machine and resume.

### Bug Incidence

Always

### nvidia-bug-report.log.gz

[nvidia-bug-report.log.gz](https://github.com/user-attachments/files/29000907/nvidia-bug-report.log.gz)

runiter:
I second this

khalangy:
Same DIFR deadlock on Ada mobile (10de:28e0), not limited to Blackwell

I'm hitting what looks like the same nvDIFRPrefetchSurfaces deadlock, but on an Ada mobile GPU rather than Blackwell, so the issue may be broader than the original report suggests.

Hardware: ASUS ROG Zephyrus G14 (GA403UV), GeForce RTX 4060 Max-Q / Mobile [10de:28e0] (Ada, AD107M) + AMD Radeon 780M iGPU (hybrid)
Driver: 580.173.02 (open kernel module)
Kernel: 7.1.3-cachyos1.fc43.x86_64 (Fedora 43)
[...]

PR #1286, opened 2026-08-11
Fix system hang from DIFR prefetch GPFIFO exhaustion

runiter:
Two related fixes for a deterministic system hang in the DIFR prefetch path. The first stops a stalled channel from taking down the machine; the second stops the channel getting into that state.

Context and the full per-boot data are in #1205.

## 1. `nvidia-push`: bound the wait for a free GPFIFO entry

`nvWriteGpEntry()` waits for a free GPFIFO entry here:

```c
// Wait for a free entry in the buffer
while (nextGpPut == ReadGpGetOffset(push_buffer)) {
    if (nvPushCheckChannelError(push_buffer)) {
        nvAssert(!"A channel error occurred in nvWriteGpEntry()");
        return FALSE;
    }
}
```

No deadline, and the only exit is `nvPushCheckChannelError()`, which returns TRUE only once RM has written `0xFFFF` into the error notifier. A channel that stalls **without faulting** — never serviced, no RC event, no notifier write — leaves that check returning FALSE forever. It is also a busy spin with no yield, so the calling kernel thread is lost and the watchdog escalates to a system hang.

Every other wait in this file is bounded: `IdleChannel()` takes a `timeoutMSec`, and the notifier wait has short and long timeouts plus an `nvPushImportYield()`. This loop is the exception.

The change gives it a deadline using `IdleChannel()`'s idiom, honours the existing `noTimeout` opt-out, yields between polls, and returns `FALSE` on expiry — a path `Kickoff()` already handles by leaving `putOffset` unchanged. No new helpers or constants.

I chose `SHORT_TIMEOUT` (3s) over `LONG_TIMEOUT` (10s) deliberately: the soft-lockup watchdog reports at ~20s, and I wanted this to surface as a logged error rather than a lockup splat. Happy to switch it for consistency with the other waits.

## 2. `nvkms-difr`: reset the prefetch channel after a CE fault

This is what makes the channel stall in the first place.

`PrefetchSingleSurface()` kicks off a copy and waits `DIFR_PREFETCH_WAIT_PERIOD_US` (10ms) for the semaphore. On expiry it returns `FAIL_CE_HW_ERROR` — but the GPFIFO entries it already wrote stay queued on a channel the CE never drained. Nothing reclaims them, so GET stops advancing and the channel permanently loses one kickoff's worth of ring.

That ring is small: `pushBufferSizeInBytes = 1024` gives `numGpFifoEntries = 16` (`nvidia-push-init.c`), 2 entries per kickoff, and `InitGpFifoExtendedBase()` consumes 2 more at alloc on Hopper+ without a matching progress-tracker release. So there is very little headroom, and a handful of faults exhausts it — after which the next kickoff waits on a ring that can never drain, which is fix 1's loop.

The change resets the channel on exactly that status, reusing the existing `Free`/`Alloc` helpers, so a fault costs nothing permanent and a later prefetch starts with GET and PUT in sync. Only `FAIL_CE_HW_ERROR` can leak — the other two failure codes return before any kickoff. If reallocation itself fails, that is remembered and `FAIL_INSUFFICIENT_L2_SIZE` is returned from then on (the code which, per the existing comment, tells RM and PMU to stop requesting prefetches) rather than leaving a freed channel for the next prefetch to dereference.

I did **not** attempt to make DIFR retry within a boot cycle. `subdeviceCtrlCmdLpwrDifrCtrl_IMPL` has no implementation in the open tree, so the "disable until next driver load" decision after a CE fault is yours, not NVKMS's, and I did not want to guess at it. In practice a resume re-enables DIFR anyway, so with this fix the next cycle gets a clean channel and can succeed if the stall was transient.

## How I hit this

RTX 5070 Ti, Ubuntu 24.04, kernel 6.17.0-1028-oem, driver 595.71.05-open, GNOME Wayland. The machine hard-hangs on **exactly the 6th suspend/resume cycle of every boot** — 7 boots out of 7, on both `deep` (S3) and `s2idle`, never once surviving a 6th. Always:

```
watchdog: BUG: soft lockup - CPU#1 stuck for 26s! [nvidia-modeset/]
RIP: nvWriteGpEntry+0xf9
  nvPushKickoff
  PrefetchHelperSurfaceEvo
  nvDIFRPrefetchSurfaces
  DifrPrefetchEventDeferredWork
  nvkms_kthread_q_callback
```

The fixed count is what pointed at a small ring being consumed one kickoff per resume rather than anything probabilistic.

Worth noting the user-visible consequence of the leak beyond the hang: because a CE fault disables DIFR until the next driver load, and each resume re-enables it for exactly one more doomed attempt, DIFR appears to be silently non-functional on this machine after the first suspend of any boot — while still accumulating the damage that eventually hangs it.

#1205 also has a report of the same deadlock signature triggered by plain display idle with no S3 involved (RTX 4060 Max-Q, 580.173.02), which fits: any repeated prefetch failure fills the ring, and suspend/resume is just a reliable way to produce one.

## Testing

**Both patches are derived from source analysis and have been compile-tested only.** They build clean against 6.17.0-1028-oem with no new warnings. I have not run them on hardware, so I can't claim they resolve the hang — only that fix 1 removes the code path's ability to spin forever, and fix 2 removes the accumulation that leads there.

Two review questions I can't answer from outside:

1. `nvPushImportYield()` has precedent in `nvidia-push.c`, but you know better than I do whether every `Kickoff()` caller is in a context where yielding is legal. Happy to gate or drop the yield and keep only the deadline.
2. Is channel alloc/free legal from `DifrPrefetchEventDeferredWork()`'s kthread-queue context? `nvkms-difr.c` already makes RM control calls from timer and kthread callbacks, but full channel reallocation from there is an assumption on my part. If it isn't safe, the reset would need deferring to a worker.

Both files are byte-identical between 595.71.05, 595.91.07 and 610.57.04, so nothing released contains a fix. I have a 100% reproducible case in six suspend cycles and am glad to build and test any alternative patch you'd prefer, on either branch.

### DIFF
--- a/src/common/unix/nvidia-push/src/nvidia-push.c
+++ b/src/common/unix/nvidia-push/src/nvidia-push.c
@@ -349,6 +349,7 @@ static NvBool nvWriteGpEntry(
     NvU32 *gpPointer;
     const NvU32 entriesNeeded = NV_PUSH_NUM_GPFIFO_ENTRIES_PER_KICKOFF;
     NvPushDevicePtr pDevice = push_buffer->pDevice;
+    NvU64 baseTime, currentTime;
 
     FillGpEntry(&push_buffer->main, putOffset, &gpEntry0, &gpEntry1);
 
@@ -358,12 +359,32 @@ static NvBool nvWriteGpEntry(
 
     nvAssert((nextGpPut % 2) == 0);
 
-    // Wait for a free entry in the buffer
-    while (nextGpPut == ReadGpGetOffset(push_buffer)) {
+    /*
+     * Wait for a free entry in the buffer.
+     *
+     * Bail out if the channel faults, but also if GET simply stops
+     * advancing: a channel that is never serviced does not necessarily
+     * raise an error notifier, and without a deadline this loop would spin
+     * in kernel context indefinitely.
+     */
+    for (baseTime = currentTime = nvPushImportGetMilliSeconds(pDevice);
+         nextGpPut == ReadGpGetOffset(push_buffer);
+         currentTime = nvPushImportGetMilliSeconds(pDevice)) {
+
         if (nvPushCheckChannelError(push_buffer)) {
             nvAssert(!"A channel error occurred in nvWriteGpEntry()");
             return FALSE;
         }
+
+        if (!push_buffer->noTimeout &&
+            (currentTime > (baseTime + NV_PUSH_NOTIFIER_SHORT_TIMEOUT))) {
+            nvPushImportLogError(pDevice,
+                "Timed out waiting for a free GPFIFO entry.");
+            nvAssert(!"Timed out waiting for a free GPFIFO entry");
+            return FALSE;
+        }
+
+        nvPushImportYield(pDevice);
     }
     gpPointer[0] = gpEntry0;
     gpPointer[1] = gpEntry1;
--- a/src/nvidia-modeset/src/nvkms-difr.c
+++ b/src/nvidia-modeset/src/nvkms-difr.c
@@ -144,6 +144,9 @@ typedef struct _NVDIFRStateEvoRec {
     /* Copy engine instance for DIFR prefetches. */
     NvU32 prefetchEngine;
 
+    /* Set if the prefetch channel could not be reallocated after a fault. */
+    NvBool channelBroken;
+
     /* For tracking which surfaces have been prefetched already. */
     NvU32 prefetchPass;
 } NVDIFRStateEvoRec;
@@ -162,6 +165,7 @@ static NvBool AllocDIFRPushChannel(NVDIFRStateEvoPtr pDifr);
 static void FreeDIFRPushChannel(NVDIFRStateEvoPtr pDifr);
 static NvBool AllocDIFRCopyEngine(NVDIFRStateEvoPtr pDifr);
 static void FreeDIFRCopyEngine(NVDIFRStateEvoPtr pDifr);
+static NvBool ResetDIFRPushChannel(NVDIFRStateEvoPtr pDifr);
 
 static NvU32 PrefetchSingleSurface(NVDIFRStateEvoPtr pDifr,
                                    NVDIFRPrefetchParams *pParams,
@@ -298,6 +302,15 @@ NvU32 nvDIFRPrefetchSurfaces(NVDIFRStateEvoPtr pDifr, size_t l2CacheSize)
         return NV2080_CTRL_LPWR_DIFR_PREFETCH_FAIL_OS_FLIPS_ENABLED;
     }
 
+    /*
+     * We no longer have a usable prefetch channel. As above, despite its
+     * wording this is the code that tells RM (and further PMU) to stop
+     * requesting prefetches until the next modeset.
+     */
+    if (pDifr->channelBroken) {
+        return NV2080_CTRL_LPWR_DIFR_PREFETCH_FAIL_INSUFFICIENT_L2_SIZE;
+    }
+
     status = NV2080_CTRL_LPWR_DIFR_PREFETCH_SUCCESS;
 
     pSubDev = &pDevEvo->gpus[0];
@@ -372,6 +385,22 @@ NvU32 nvDIFRPrefetchSurfaces(NVDIFRStateEvoPtr pDifr, size_t l2CacheSize)
     }
 
 out:
+    /*
+     * A CE error means we kicked off a prefetch that was never consumed,
+     * leaving GPFIFO entries queued on the channel for good. Reset the
+     * channel so that the failure doesn't cost us ring space permanently:
+     * otherwise repeated failures eventually exhaust the GPFIFO and the
+     * next kickoff has nothing left to wait for.
+     */
+    if (status == NV2080_CTRL_LPWR_DIFR_PREFETCH_FAIL_CE_HW_ERROR) {
+        if (!ResetDIFRPushChannel(pDifr)) {
+            pDifr->channelBroken = TRUE;
+
+            nvEvoLogDev(pDevEvo, EVO_LOG_WARN,
+                        "Failed to reset the DIFR prefetch channel.");
+        }
+    }
+
     return status;
 }
 
@@ -431,6 +460,31 @@ static NvBool AllocDIFRPushChannel(NVDIFRStateEvoPtr pDifr)
     }
 
     if (!nvPushAllocChannel(&params, &pDifr->prefetchPushChannel)) {
+        return FALSE;
+    }
+
+    return TRUE;
+}
+
+/*
+ * Tear down and reallocate the prefetch channel.
+ *
+ * A prefetch that fails with a CE error has already written GPFIFO entries
+ * that the copy engine never consumed. Nothing else reclaims them, so GET
+ * stays where it is and the channel permanently loses that much of its
+ * (small) GPFIFO ring. Reallocating the channel puts GET and PUT back in
+ * sync so a later prefetch can start from a clean slate.
+ */
+static NvBool ResetDIFRPushChannel(NVDIFRStateEvoPtr pDifr)
+{
+    FreeDIFRCopyEngine(pDifr);
+    FreeDIFRPushChannel(pDifr);
+
+    if (!AllocDIFRPushChannel(pDifr) ||
+        !AllocDIFRCopyEngine(pDifr)) {
+        FreeDIFRCopyEngine(pDifr);
+        FreeDIFRPushChannel(pDifr);
+
         return FALSE;
     }
 


runiter:
Good news, my proposed fix at PR below is working well in my own pc. No more hangs!

For those who like to try it, I built this easy script in attached zip file:
[nvidia-difr-hang-fix.zip](https://github.com/user-attachments/files/31073466/nvidia-difr-hang-fix.zip)


Simply run these commands:

```
./build.sh
sudo ./install.sh 
sudo reboot
```

Cartagines:
Independent reproduction on a different GPU generation, compositor and trigger,
  on 610.57.04. Adding it here since you note the files are byte-identical across
  595.71.05 / 595.91.07 / 610.57.04 — this confirms in the field that the newest
  release still hangs.

  ## Configuration

  | | |
  |---|---|
  | GPU | RTX 4070 Max-Q / Mobile, AD106M `[10de:2860]` (**Ada**, not Blackwell) |
  | Driver | `nvidia-open` **610.57.04** |
  | Kernel | 7.1.8-arch1-3 (Arch) |
  | Session | **KWin 6.7.4 / Plasma 6**, Wayland (not GNOME) |
  | Display | **eDP-1 hangs off the dGPU** (`card1-eDP-1`); the iGPU exposes no eDP connector |

  ## Trigger: display idle, with no suspend anywhere in the picture

  You mention a second-hand report of the non-S3 case in #1205. This is a
  first-hand one, and the journal is unambiguous: **zero `PM: suspend entry`
  events across the entire retained history.** This machine has never suspended.
  The hang happens with the session sitting idle on the lock screen.

  Four occurrences with an identical signature — 2026-08-06 18:09, 08-12 18:36,
  08-13 19:56, 08-17 12:02 — each ending in a forced power-off. The nine boots
  before 08-06 all shut down cleanly.

  ## Stacks

  Holder, spinning in `R` exactly as in your report:

  ```
  task:nvidia-modeset/ state:R  running task
    ? nvWriteGpEntry            [nvidia_modeset]
    ? nvPushKickoff             [nvidia_modeset]
    ? PrefetchHelperSurfaceEvo  [nvidia_modeset]
    ? nvDIFRPrefetchSurfaces    [nvidia_modeset]
    ? DifrPrefetchEventDeferredWork [nvidia_modeset]
    ? nvkms_kthread_q_callback  [nvidia_modeset]
  ```

  Waiter — and this is a **different entry point** from the atomic-modeset path
  in the other reports:

  ```
  INFO: task kwin_wayland:1343 blocked for more than 122 seconds.
  INFO: task kwin_wayland:1343 blocked on a semaphore likely last held by task nvidia-modeset/:350
  task:kwin_wayland    state:D
    down+0x53
    nvkms_ioctl_from_kapi+0x58     [nvidia_modeset]
    CreateSurface+0x214            [nvidia_modeset]
    nv_drm_framebuffer_create+0x365 [nvidia_drm]
    drm_internal_framebuffer_create
    drm_mode_addfb2
  ```
  
  So the deadlock is reachable through plain `ADDFB2`, via
  `nvkms_ioctl_from_kapi()` (not the `..._try_pmlock` variant), which takes
  `nvkms_lock` with an uninterruptible `down()`. That makes every blocked DRM
  client unkillable — `SIGKILL` included — so there is no recovery short of a
  hard reset.

  Two details that support your "stall without faulting" reading:
  
  - **No Xid, in any of the four occurrences.** Consistent with
    `nvPushCheckChannelError()` never firing because RM never wrote `0xFFFF`.
  - The kernel stays fully alive while the display is dead. In the 08-17 hang,
    an application completed an HTTPS request to a remote host **39 minutes
    after the compositor died** — DNS, TCP, TLS and journald writes all working.
    Only the graphics path is gone.

  ## Re: your first review question (is the yield legal?)

  > `nvPushImportYield()` has precedent in `nvidia-push.c`, but you know better
  > than I do whether every `Kickoff()` caller is in a context where yielding is
  > legal.

  For what it's worth from outside as well, this looks safe in the published
  tree, on three grounds:

  1. **nvkms is the only consumer of `nvidia-push` here.** There is exactly one
     imports table — `src/nvidia-modeset/src/nvkms-push.c:246`. No other host
     driver in this tree can reach `Kickoff()` from an atomic context.

  2. **The DIFR path is already in sleeping context.**
     `nvkms_kthread_q_callback()` takes `nvkms_read_lock_pm_lock()` (line 1021)
     and then `down(&nvkms_lock)` (line 1023) before invoking `timer->proc()`.
     Both are sleeping locks, and `nvkms_yield()` is just `schedule()`
     (`nvidia-modeset-linux.c:735`).

  3. **Direct precedent for the same operation.** `nvkms-dma.c` waits for the
     GPU to drain an EVO channel in `nvEvoPollForEmptyChannel()` and calls
     `nvkms_yield()` unconditionally in that loop (line 85), with
     `nvEvoMakeRoom()` bounding it at 5 s (line 97). Same semantics as this
     loop, bounded and yielding.

  Only the two channels allocated through `nvPushAllocChannel()` are affected —
  DIFR and HeadSurface-3D — so the blast radius of the yield is small.

  ## Re: the timeout value

  > Happy to switch it for consistency with the other waits.

  One data point for whichever you pick. The legitimate worst case on the DIFR
  channel is bounded by the ring depth, not by how long the copy takes:

  ```
  pushBufferSizeInBytes = 1024                       nvkms-difr.c:422
  numGpFifoEntries      = 1024 / 64 = 16             nvidia-push-init.c:1285
                        / 2 per kickoff  =  8 kickoffs in flight
  ```

  and `PrefetchSingleSurface()` documents each copy at a few hundred µs, "slightly
  less than 2 milliseconds for a single 4k display". So ~16 ms is the realistic
  ceiling for a healthy channel, and `SHORT_TIMEOUT` already carries ~190× margin.
  HeadSurface's ring is 2048 entries, so filling it means a wedged GPU rather than
  a loaded one either way.

  ## Testing offer
  
  I arrived at the same fix for the first hunk independently before finding this
  PR, which is at least a second pair of eyes on that approach. What I have that
  may be useful to you: a **userspace regression test** for the wait loop.

  `nvidia-push` reaches the OS only through `pDevice->pImports`, and reads GP_GET
  from plain memory at `progressSemaphore.ptr[]`, so a stalled ring can be mocked
  exactly — no GPU, no kernel headers, no root. Against the unpatched loop it
  hangs; against the bounded one it passes in milliseconds, and a third case
  covers the ring freeing up mid-wait so the normal kickoff path is not
  regressed. It builds with gcc and clang, clean under ASan/UBSan.

  Happy to post it if it would help, and equally happy to run either of your
  patches on this hardware — my trigger is display idle rather than suspend, so
  it exercises a different route into the same loop.

  ## One caution on the sibling loop

  `__nvPushMakeRoom()` has the same unbounded shape, and it would be natural to
  give it the same treatment — but it returns `void`, and its callers write
  unconditionally afterwards. `nvPushHeader()` does
  `__nvPushSetMethodDataSegment()` and then `freeDwords -= (count+1)` with no
  check, so an early return there would write past the available space and
  underflow `freeDwords` (`NvU32`) into a self-amplifying overrun of a
  DMA-mapped buffer. Bounding it safely needs the failure to be propagated
  through the push macros first. Worth a comment there if this lands, so nobody
  "fixes" it the obvious way.

aritger:
This is excellent debugging, @runiter.  Thank you for the thorough analysis in https://github.com/NVIDIA/open-gpu-kernel-modules/issues/1205 and for the proposed fixes here.  I need a little time to get more familiar with this particular code.  But I wanted to acknowledge this pull request, and I'll try to answer your open questions, soon.

Cartagines:
Let me know if i can help in any way @aritger @runiter

runiter:
@aritger you're welcome. Glad to hear that the fix is being looked at seriously.
FYI the fix continues to work well in local machine. Before the fix I would get consistent hang at 6th s3 wakeup.
Now I'm on 15th s3 sleep and 2 weeks time and not a single hang so far.

@Cartagines thank you for confirming it

albanD:
In case it helps, I've been running this patch for the past 2/3 days and it fixed the freezing as well! Looking forward for it to make it into a release so I can stop re-patching the driver :D

aritger:
FYI, the 595.99.02 released today contains a related change (tear down DIFR before suspend, and realloc DIFR after resume).  I suspect that change resolves the user-visible symptom that motivated this pull request.  Could anyone test that out to confirm?

(I'd still like to include the changes in this pull request in future releases for improved general resilience.  I need to make some time to test them locally, first...)

albanD:
Given that rpm-fusion is on 610.57.04 right now, I don't expect that they'll ship this one so we will have to rebuild from source.
My plan would be to re-patch your change
```
  git diff 595.71.05..595.99.02 -- \
    src/nvidia-modeset/src/nvkms-evo.c
```
on top of my current driver 595.71.05 (that allows me to only update nvidia-modeset.ko and keep everything else the same).

Would that be an ok validation?

runiter:
@aritger I installed the official NVIDIA 595.99.02 release, which tears down the DIFR channel before suspend and reallocates it after resume. I will let you know if the hanging issue returns.

QcN3ep:
@aritger  Following up on your question about 595.99.02: I built and tested debug drivers for 610.57.04 and 595.99.02 on the same machine and with the same hardware/display configuration described below. This configuration differs from the systems previously reported in this PR, so I wanted to add the result as an independent data point.

## Configuration

| Item | Configuration |
| --- | --- |
| CPU | AMD Ryzen 9 8945HX |
| GPU | GeForce RTX 5060 Laptop GPU |
| iGPU | AMD Radeon Raphael (driving the internal eDP) |
| Display | External HDMI connected to the NVIDIA GPU |
| Base driver | `nvidia-open-dkms 610.57.04-1` |
| Kernel | `7.1.9-zen1-2-zen` |
| Session | KDE Plasma Wayland |

### Matching user-space, GSP, and kernel-module versions

- **610.57.04**
  - User-space components: `nvidia-utils 610.57.04-1`
  - GSP firmware: `/usr/lib/firmware/nvidia/610.57.04/gsp_ga10x.bin` and `gsp_tu10x.bin`
  - Kernel module: `nvidia-open-dkms 610.57.04-1`
- **595.99.02**
  - User-space components: `nvidia-utils 595.99.02-1`
  - GSP firmware: `/usr/lib/firmware/nvidia/595.99.02/gsp_ga10x.bin` and `gsp_tu10x.bin`
  - Kernel module: `nvidia-open-dkms 595.99.02-1`

## Failure Symptoms

This machine has experienced 2 complete hangs so far during normal display-idle operation. The screen remained lit on the last frame; the keyboard, mouse, and Caps Lock indicator stopped responding. The kernel continued to operate normally, but effective interaction was impossible, so I had to force a reboot.

The interval between the two failures was less than 12 hours. Journal inspection found no `PM: suspend entry/exit` events before either failure in the corresponding boot.

The key stack trace was identical in each incident:

```text
nvWriteGpEntry
nvPushKickoff
PrefetchHelperSurfaceEvo
nvDIFRPrefetchSurfaces
DifrPrefetchEventDeferredWork
nvkms_kthread_q_callback
```

Afterwards, `kwin_wayland` blocked in the NVKMS/DRM path while waiting for the same lock. No NVIDIA Xid, GPU reset, MCE, AER, or OOM event was recorded.

## Debug-Build Observations

I first built an instrumented 610.57.04 driver with additional logging and ran it with the same KDE session, display topology, and idle workload in a single-boot observation window. The logs covered approximately 163 seconds after boot. The previous complete hang did not recur during this observation window, but the build captured one naturally occurring `NV2080_CTRL_LPWR_DIFR_PREFETCH_FAIL_CE_HW_ERROR` response. The corresponding `NV2080_CTRL_CMD_LPWR_DIFR_PREFETCH_RESPONSE` control call returned success. Measured GPFIFO occupancy remained at `4`/`16` entries, and `5` post-timeout samples showed no `gp_get`, `gp_put`, or occupancy growth. A log excerpt is shown below:

```text
[  162.541021] ... DIFR_MEASURE ring event=ce_timeout seq=1 gp_get=4 gp_put=8 gp_next=10 occupancy=4 entries=16
[  162.546021] ... DIFR_MEASURE response status=4 success=1
[  163.067073] ... DIFR_MEASURE ring event=post_timeout seq=5 gp_get=4 gp_put=8 gp_next=10 occupancy=4 entries=16
```

I then built the corresponding 595.99.02 driver with additional logging and ran it in the same environment in a single-boot observation window. The logs covered approximately 355 seconds after boot. No new complete hang occurred during this observation window, and the captured CE-fault sample likewise showed no ring growth. A log excerpt is shown below:

```text
[  354.210021] ... DIFR_MEASURE ring event=ce_timeout seq=1 gp_get=6 gp_put=10 gp_next=12 occupancy=4 entries=16
[  354.213020] ... DIFR_MEASURE response status=4 success=1
[  354.732063] ... DIFR_MEASURE ring event=post_timeout seq=5 gp_get=6 gp_put=10 gp_next=12 occupancy=4 entries=16
```

After each of these two tests, I left the machine idle for 15 hours to simulate normal day-to-day use. No hang occurred, so the issue was not reproduced.

Because naturally occurring CE events are rare and the measurement builds add extra diagnostics, this is not a decisive runtime A/B result. In particular, the absence of another hang does not prove that 595.99.02 fixes the path that does not involve suspend.

## Source Comparison

Comparing the open kernel-module sources identifies the coverage boundary of the 595.99.02 change:

- 595.99.02 adds `AllocDIFRState()` and `FreeDIFRState()` in `nvkms-evo.c`. DIFR state is freed before suspend and allocated again after resume.
- The 610.57.04 release does not contain these suspend/resume DIFR lifecycle calls.
- `nvkms-difr.c`, `nvidia-push.c`, and the DIFR prefetch event path are unchanged between the two release tags. Therefore, the CE-fault handling and the unbounded `nvWriteGpEntry()` wait in the open tree are unchanged as well.
- Reinitializing the DIFR state also reinitializes its push channel, so the 595 change can clear stale channel state during an actual suspend/resume cycle.

However, the relevant DIFR prefetch path is the same during normal display-idle operation. `PrefetchSingleSurface()` starts a CE copy, waits 10 ms for the completion semaphore, and returns `FAIL_CE_HW_ERROR` on timeout. In the open-tree sources, neither 595.99.02 nor 610.57.04 adds channel recovery on this path. Since none of the observed hangs followed suspend/resume, the 595 lifecycle change would not run on the observed route.

## Conclusion

The runtime debug-build comparison is not decisive because naturally occurring failures are uncommon and measurement builds can change behavior. Nevertheless, the hangs observed with the production 610.57.04 driver, together with the source comparison, indicate that the 595.99.02 lifecycle change does not cover this no-suspend display-idle path. This is consistent with the independent KDE/Wayland reproductions already reported in this discussion.

PR #1286 appears to cover the remaining path by bounding the wait for a free GPFIFO entry and attempting to tear down and reallocate the DIFR prefetch channel after `CE_HW_ERROR`. I have been using this fix for some time and have not seen another hang so far.

runiter:
@aritger good news. I tried `595.99.02` and it seems like it fixed the issue! I suspended and resumed 10 times already and no hang. Normally it would always hang on the 6th resume. So looks like the issues I was having is already fixed.

I had to build version `595.99.02` locally because looks like it's not released yet. Do you know when it will be officially released?

robert-burton:
I'd been facing constant freezing incidents on my device using current Nvidia drivers, and I'm pleased to report that the proposed changes have noticeably helped on my end.

**System:**
Model: Lenovo Legion Slim 5 16ARP9
GPU: NVIDIA GeForce RTX 4060 Max-Q / Mobile (no iGPU, NVIDIA is the sole display controller)
Driver: 610.57.04 (open)
Kernel: Linux 7.2.0-1-cachyos
Desktop Environment: Gnome 50 (Wayland)

The main triggering scenarios seemed to be the following, from most to least likely:

a) Waking up from suspend (s2idle), in which the display would freeze either immediately after wake up, or a few seconds later.
b) Waking up the display after it has blanked out due to inactivity, _before_ the machine has actually entered suspend.
c) Interacting with a window (focusing, moving, resizing, etc.) during normal desktop use and light workloads.

Of the previously listed, a) and b) were overwhelmingly more common, and would encounter them almost every day I booted Linux on my laptop, sometimes more than once on the same day. 

Once frozen, the device would become completely unresponsive with the display stuck showing the last frame, but audio playback and network connectivity would remain. Trying to switch to TTY wouldn't work and I would be forced to hard reboot the machine. This applies to all previously mentioned scenarios.

With that said, I've been running the proposed changes for about 5 days and scenarios a) and b) have not triggered freezes at all.

I have, however, had two freezing incidents since I applied the patch on my install:


1. This one happened during normal desktop use as described in c), only this time I was actually able to switch to TTY and back to graphical session to undo the freeze, after which the device appeared to continue working normally.


2. This one was more severe, and followed a previously unseen template. The machine was left idle for about 30 minutes, more than enough time for it to enter suspend (currently have suspend set to 15 min). After coming back to it the display was blank (expected), but the power button LED was not flashing and the keyboard backlight was still on (not expected). So in this case the freezing occurred after the screen blanked out due to inactivity, but before entering suspend. This is different from scenario b) in that the freeze took place when the screen was blanked out, while in b) it was the _waking_ of the display that triggered it. In fact, something that makes this freezing event stand out is that, unlike all other scenarios described, this one happened on its own, without any form of user interaction, which is an element that's present in all previously described cases (waking up the device, interacting with windows, etc.)
   
I was unable to recover any useful logs from the last incident, but will try to do so if it happens again. Another maybe useful data point: both of these incidents took place somewhat closely, it may have been in the same 24h period (without rebooting), about two days ago.

To summarize: a) and b) seem completely fixed, c) was not fully fixed but at least I'm able to get my session back, and a previously unseen freeze happened in the meantime, which may be related to the proposed changes but I can't say for certain. Will share more info if I manage to get it.

robert-burton:
Quick followup on my previous post: I've been running the patch daily since I last posted and I've seen no freezing on my machine since then. No particular changes to the system on my end, I just kept updating and using my laptop as usual. It's been a pretty clean run so far, with only two freezing incidents in the first few days (as described previously), which I haven't been able to replicate through normal usage.

Looking forward to the proposed changes being merged soon, they've helped immensely on my end.

aritger:
Thanks for your testing, @runiter.

595.99.02 was released on August 27:
    https://www.nvidia.com/en-us/drivers/details/278285/
    https://github.com/NVIDIA/open-gpu-kernel-modules/releases/tag/595.99.02

615.71.09, released today, also contains the same change.  So, that may be easier for others to test than to roll back to 595.xx:
    https://www.nvidia.com/en-us/drivers/details/278450/
    https://github.com/NVIDIA/open-gpu-kernel-modules/releases/tag/615.71.09

PR #1346, opened 2026-09-09
fix: correct conditional rwlock acquisition results

bgly:
Linux documents [down_read_trylock()](https://github.com/torvalds/linux/blob/v6.17/kernel/locking/rwsem.c#L1571-L1582) and [down_write_trylock()](https://github.com/torvalds/linux/blob/v6.17/kernel/locking/rwsem.c#L1613-L1625) as returning 1 on successful acquisition and 0 on contention.

Negate both checks so the conditional rwlock wrappers return `NV_ERR_TIMEOUT_RETRY` only when acquisition fails, and `NV_OK` when it succeeds.


aaronp24:
Thanks for reporting this. It looks like this was already fixed internally and the fix will be in a future release.

